PR #7, opened 2018-10-11
Disabled Automatic Setup of Logging

EdsterG:
Libraries should never configure logging on import. Doing so will destroy the configuration of the calling library, which is not desirable.


hungpham2511:
@EdsterG Thanks for the suggestion, that's reasonable. Can you remove the line altogether? after that I will happily merge the pr.

EdsterG:
Will do, just wanted to make sure no one added the call back in some time in the future. =)

PR #21, opened 2018-11-07
Change how solver wrappers are configured and Implement Automatic Scaling

hungpham2511:
Some changes:
- Each solver now enforces its own limits on x, u and so on. This is to ensure numerical stability for solvers like ECOS or Seidel.
- A bug fixed in Seidel.
- Test cases reorganized
- Implement problem scaling, accessible via argument parameter `scaling`. Auto-scaling is available, used by setting `scaling` to 1.


hungpham2511:
Seidel fails to solve the following instance at difference scaling levels. Tested at 1e-1, 1e-2 and 1e-3.

``` python
import toppra as ta
import toppra.constraint as constraint
import toppra.algorithm as algo
import numpy as np
import matplotlib.pyplot as plt

ta.setup_logging("DEBUG")


def main():
    # waypts = [[0], [5], [10]]
    # path = ta.SplineInterpolator([0, 0.5, 1.0], waypts)
    waypts = [[0], [1e-8], [0]]
    path = ta.SplineInterpolator([0, 0.5, 1.0], waypts)
    # NOTE: When constructing a path, you must "align" the waypoint
    # properly yourself. For instance, if the waypoints are [0, 1, 10]
    # like in the above example, the path position should be aligned
    # like [0, 0.1, 1.0]. If this is not done, the CubicSpline
    # Interpolator might result undesirable oscillating paths!
    vlim = np.array([[-10, 10]])
    alim = np.array([[-4, 4]])
    pc_vel = constraint.JointVelocityConstraint(vlim)
    pc_acc = constraint.JointAccelerationConstraint(
        alim, discretization_scheme=constraint.DiscretizationType.Interpolation)

    instance = algo.TOPPRA([pc_vel, pc_acc], path, solver_wrapper='seidel',
                           gridpoints=np.linspace(0, 1.0, 1001), scaling=1e-3)
    jnt_traj, aux_traj, data = instance.compute_trajectory(0, 0, return_data=True)
```

hungpham2511:
For pathological cases, like path with minimal/near zero motions, seidel solver wrapper works badly. Probably because it does not scale the Linear Program prior to solving. Since reimplementing seidel does not justify the benefit, as one can always switch to a more stable solver wrapper such as qpOASES for these cases, we no longer require seidel to pass tests with very small motions.

PR #23, opened 2018-11-27
add compute_reachable_sets()

herjia:
In reachability_algorithm.py, I have added two functions, which are named _one_step_forward() and compute_reachable_sets() respectively. With these function, we can compute the reachable sets.

### DIFF
--- a/toppra/algorithm/reachabilitybased/reachability_algorithm.py
+++ b/toppra/algorithm/reachabilitybased/reachability_algorithm.py
@@ -338,3 +338,51 @@ class ReachabilityAlgorithm(ParameterizationAlgorithm):
             return sdd_vec, sd_vec, v_vec, K
         else:
             return sdd_vec, sd_vec, v_vec
+
+
+    def _one_step_forward(self, i, L_current, feasible_set_next):
+        res = np.zeros(2)
+        if np.isnan(L_current).any() or i < 0 or i > self._N:
+            res[:] = np.nan
+            return res
+        nV = self.solver_wrapper.get_no_vars()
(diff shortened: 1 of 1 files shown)


toppra/algorithm/reachabilitybased/reachability_algorithm.py
```diff
@@ -338,3 +338,51 @@ def compute_parameterization(self, sd_start, sd_end, return_data=False):
             return sdd_vec, sd_vec, v_vec, K
         else:
             return sdd_vec, sd_vec, v_vec
+
+
+    def _one_step_forward(self, i, L_current, feasible_set_next):
```
hungpham2511: This function's name should be `_reach` to maintain consistency with the paper. :+1:

toppra/algorithm/reachabilitybased/reachability_algorithm.py
```diff
@@ -338,3 +338,51 @@ def compute_parameterization(self, sd_start, sd_end, return_data=False):
...
+        x_opt_0 = opt_0[1]
+        u_opt_0 = opt_0[0]
+        x_lower = x_opt_0 + 2 * deltas * u_opt_0
+
+        res[:] = [x_lower, x_upper]
+        return res
+
+    def compute_reachable_sets(self, sdmin, sdmax):
+        assert sdmin <= sdmax and 0 <= sdmin
+        feasible_sets = self.compute_feasible_sets()
```
hungpham2511: I believe we can skip computing the feasible sets. Let's discuss that tomorrow!

Issue #82, opened 2020-04-13
C++ Implementation


hungpham2511:
We will use this issue to flesh out some key designs of the C++ interface.

There are a few core components that are required.
- [x] Trajectory representation;
- [x] Constraint representation;
- [x] LP solver wrapper;
- [x] Core TOPPRA pipeline;
- [x] Matrix manipulation libraries in C++?

I am most inexperienced with trajectory representation in C++. Use OpenRAVE before with its representation, but bringing it over is a huge dependency that I am not willing to make. In the Python implementation I am basically using `scipy.interpolate` extensively.

jmirabel:
One extra mention: [Pinocchio](https://github.com/stack-of-tasks/pinocchio) could also be used to interpolate the trajectories. It does the interpolation in Lie groups.

hungpham2511:
Totally agree on the abstract base class. I propose we make use of the current Python interfaces. Have a look:
- constraints: https://github.com/hungpham2511/toppra/blob/c013302cf78fe369c61a293789ee32cac0f4f759/toppra/constraint/linear_constraint.py#L7
- trajectory/geometric path:
https://github.com/hungpham2511/toppra/blob/c013302cf78fe369c61a293789ee32cac0f4f759/toppra/interpolator.py#L84
- LP/QP solver wrapper interface
https://github.com/hungpham2511/toppra/blob/c013302cf78fe369c61a293789ee32cac0f4f759/toppra/solverwrapper/solverwrapper.py#L42

The main toppra algorithm is relatively straightforward once we have all the pieces working.

Regarding license, I propose to use the MIT license, the same as the Python codebase. What do you think?

Another point is testing, I would like the code to have good unit test coverage. Shall we use google-test?

jmirabel:
> * constraints: https://github.com/hungpham2511/toppra/blob/c013302cf78fe369c61a293789ee32cac0f4f759/toppra/constraint/linear_constraint.py#L7

Any reason not to use this one:
https://github.com/hungpham2511/toppra/blob/c013302cf78fe369c61a293789ee32cac0f4f759/toppra/constraint/constraint.py#L30


> Regarding license, I propose to use the MIT license, the same as the Python codebase. What do you think?

MIT is fine for me.

> Another point is testing, I would like the code to have good unit test coverage. Shall we use google-test?

I almost never used google-test but I guess it isn't hard to use.

hungpham2511:
[...]

PR #83, opened 2020-04-15
First draft of LinearConstraint

jmirabel:
Related to #82

### DIFF
--- a/cpp/src/toppra/constraint.cpp
+++ b/cpp/src/toppra/constraint.cpp
@@ -27,7 +27,9 @@ void LinearConstraint::computeParams(const GeometricPath& path, const Vector& gr
     std::invalid_argument("Wrong number of b vectors");
   if (c.size() != N_1)
     std::invalid_argument("Wrong number of c vectors");
-  if (F.size() != N_1)
+  if (constantF() && F.size() != 1)
+    std::invalid_argument("Expected only one F matrix");
+  if (!constantF() && F.size() != N_1)
     std::invalid_argument("Wrong number of F matrices");
   if (g.size() != N_1)
     std::invalid_argument("Wrong number of g vectors");
@@ -41,8 +43,12 @@ void LinearConstraint::computeParams(const GeometricPath& path, const Vector& gr
       std::invalid_argument("Wrong b[i] vector size.");
     if (c[i].size() != m)
       std::invalid_argument("Wrong c[i] vector size.");
-    if (F[i].rows() != k || F[i].cols() != m)
-      std::invalid_argument("Wrong F[i] matrix dimensions.");
+    if (constantF())
+      if (i == 0 && (F[i].rows() != k || F[i].cols() != m))
+        std::invalid_argument("Wrong F[0] matrix dimensions.");
+    else
+      if (F[i].rows() != k || F[i].cols() != m)
+        std::invalid_argument("Wrong F[i] matrix dimensions.");
     if (g[i].size() != k)
       std::invalid_argument("Wrong g[i] vector size.");
   }
--- a/cpp/src/toppra/constraint.cpp
+++ b/cpp/src/toppra/constraint.cpp
@@ -4,7 +4,11 @@ namespace toppra {
 
 std::ostream& LinearConstraint::print(std::ostream& os) const
 {
-  return os << "    Discretization Scheme: " << discretizationType_ << "\n";
+  return os <<
+    "    Discretization Scheme: " << discretizationType_ << "\n"
+    "    Has " << (hasLinearInequalities() ? "linear inequalities, " : "")
+    << (hasUbounds() ? "bounds on u, " : "")
+    << (hasXbounds() ? "bounds on x, " : "") << "\n";
 }
 
 void LinearConstraint::discretizationType (DiscretizationType type)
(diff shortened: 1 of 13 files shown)


hungpham2511:
This is more like a question rather comment. Would you suggest to implement the library as header-only, or the more traditional source/header style?

cpp/src/toppra/constraint.cpp
```diff
@@ -0,0 +1,43 @@
...
+    Matrix& a, Matrix& b, Matrix& c, Matrix& F, Matrix& g, Bounds& ubound, Bounds& xbound)
+{
+  Eigen::Index N_1 = gridpoints.size(),
+               m = a.cols(),
+               k = g.cols();
+  if (b.cols() != m)
+    std::invalid_argument("Wrong number of columns of matrix b");
+  if (c.cols() != m)
+    std::invalid_argument("Wrong number of columns of matrix c");
+  //if (F.cols() != m)
```
hungpham2511, line 29: We will need F to represent the constraint. Why is it commented out?

jmirabel: Because F is not of the good type yet. It is a tensor. So either we rely on the Eigen Tensor module or we use an internal type.
At the moment, the constraint compute the parameters of all the gridpoints at once. I wonder if it would make sense to compute the parameter for on point only.

The pros are:
- this would make it easier to handle constraints only on a subinterval,
- we would not need tensors.

jmirabel:
Header only libraries are useful for library that provides feature through templetization:
- CRTP like libraries: I don't think there will be any benefits for this library and the code would be much harder to write,
- scalar templetization: this could make sense but I think there is a smarter way of getting the derivative, if needed.

cpp/src/toppra/constraint.cpp
hungpham2511: Ah, right I miss that point. Actually we do not use the fact that F is a tensor, so I think it's find letting F be a vector of matrices.

On a related note, probably it is more reasonable to have a,b,c, g be vectors of 1d-matrices as well.

hungpham2511:
> Header only libraries are useful for library that provides feature through templetization:
> 
> * CRTP like libraries: I don't think there will be any benefits for this library and the code would be much harder to write,
> * scalar templetization: this could make sense but I think there is a smarter way of getting the derivative, if needed.

Right, thanks for the suggestion. In that case let's do it the conventional way then.

cpp/src/toppra/constraint.cpp
jmirabel: Ok. One question while we are at on this point. I have scenarios where it would make sense to have certain constraints only on a sub-interval. Do you think we can allow this ?

hungpham2511: From the algorithmic viewpoint, it's perfectly fine. You only need to set F[i] for i outside of the interested range to be zero, and g[i] to -1. 

From a physical perspective, however, the discontinuity can be worrying. You would probably want to slowly relax the constraint instead of abruptly set the coefficients to these values.

jmirabel:
> * scalar templetization: this could make sense but I think there is a smarter way of getting the derivative, if needed.

One question on this point: at each grid point, when the best control is selected, it is possible to know what constraint are active for this control, isn't it ? do you know how hard it would be to reconstruct the set of active constraints ?

hungpham2511:
> > * scalar templetization: this could make sense but I think there is a smarter way of getting the derivative, if needed.
> 
> One question on this point: at each grid point, when the best control is selected, it is possible to know what constraint are active for this control, isn't it ? do you know how hard it would be to reconstruct the set of active constraints ?

I would say it should be very easy. qpOASES or any active set solvers already give you what are the active constraints.

cpp/src/toppra/constraint.cpp
jmirabel: Won't the abrupt change of constraint be smoothed by the forward and backward pass ? If the constraint you add is a limit of the cartesian velocity of the end-effector, the end-effector velocity should *smoothly* decrease a few step ahead the constraint change.

jmirabel:
I updated the code according to our discussion.

jmirabel:
I started looking at the various constraint implementation and I am thinking of changing the design. Please wait before merging.

cpp/src/toppra/constraint.cpp
hungpham2511: Yes, mathematically it will work. The robot will slow down just nice to satisfy other constraints. 

However, you never until you try the code on a real, physical robot.

jmirabel:
I am now satisfied with the design. Comments are welcome.

cpp/src/toppra/constraint.cpp
```diff
@@ -0,0 +1,153 @@
...
+  }
+
+  computeParams_impl(path, gridpoints, a, b, c, F, g);
+  if (discretizationType_ == Interpolation)
+    collocationToInterpolate(gridpoints, a, b, c, F, g);
+}
+
+void LinearConstraint::collocationToInterpolate (const Vector& gridpoints,
+    Vectors& a, Vectors& b, Vectors& c,
+    Matrices& F, Vectors& g)
```
jmirabel, line 61: This does the same this as function `canlinear_collocation_to_interpolate`. I wrote that function by understanding the Python code. We should double check it later. As of now, it is hard to unit-test it because we do have a mean of generating relevant values.

cpp/src/toppra/constraint.cpp
```diff
@@ -0,0 +1,160 @@
...
+      if (F[i].rows() != k_ || F[i].cols() != m_)
+        throw std::invalid_argument("Wrong F[i] matrix dimensions.");
+      if (g[i].size() != k_)
+        throw std::invalid_argument("Wrong g[i] vector size.");
+    }
+  }
+
+  computeParams_impl(path, gridpoints, a, b, c, F, g);
+  if (discretizationType_ == Interpolation)
+    collocationToInterpolate(gridpoints, a, b, c, F, g);
```
jmirabel, line 56: In the current design, the output matrices are resized when the discretization type is Interpolation.

cpp/src/toppra/constraint.hpp
```diff
@@ -1,6 +1,213 @@
...
+ *  \f{eqnarray}
+ *      \mathbf a_i &= (1, -1, 0, 0).T \\
+ *      \mathbf b_i &= (0, 0, 1, -1).T \\
+ *      \mathbf c_i &= (-u^b_{i,1}, u^b_{i,0}, -x^b_{i,1}, x^b_{i,0}).T \\
+ *      \mathbf F   &= I_4 \\
+ *      \mathbf g_i &= 0_4 \\
+ *  \f}
+ *
+ * */
+class BoxConstraint : public LinearConstraint {
```
hungpham2511, line 148: I understand why you want to have BoxConstraint, it's true that xbound and ubound can both be formulated in the general form as a Linear constraint.

I don't agree with this design because ubound and xbound were there for the sake of efficiency. Many solvers (qpOASES included) have direct bound on its variables for efficiency reason. I want to take advantage of that. 

1. For the sake of generality, why even use BoxConstraint at all? LinearConstraint is sufficient.

cpp/src/toppra/constraint.cpp
hungpham2511: Sure. We can always check it against the Python implementation.

hungpham2511:
Thanks for the PR. I think it looks great! I only have a comment regarding the BoxConstraint sub-class.

cpp/src/toppra/constraint.hpp
jmirabel: > Many solvers (qpOASES included) have direct bound on its variables for efficiency reason.

This is exactly how I understood this and it is why the implementation of `BoxConstraints::computeParams_impl` throws an exception. What I have in mind is that wrappers of solvers which handle box constraints as a special case should do a special case about them. What I mean is, in the wrapper class, there is something like:
```c++
if (dynamic_cast<BoxConstraint*>(constraint)) {
 // treat it as a box constraint
} else {
 // treat it as a general linear constraint
}
```

If it is a problem, we can replace the dynamic_cast by a virtual function. For what it is worth, dynamic_cast makes the code looks dirtier (in my opinion) but the overhead is better than that of a virtual function (I don't think we will be optimizing until this level anyway).

hungpham2511: I agree that we don't have to optimize to that level.

What really bugs me is the fact that we have two different data structures to represent the LinearConstraint class, while having just one is sufficient. Since this is the interface between the problem definition and the algorithm, I would prefer something simple and straightforward.

That said, suppose there is a kind of constraint that actually evaluates to a combination of **both** the (a,b,c, F, g) form and (xb, ub) form, we will need to produce two constraints objects, one LinearConstraint and one BoxConstraint is that right? Yet, I don't think this is a big problem. Mostly concerned with the complexity of having two interfaces to "choose".

jmirabel: > I would prefer something simple and straightforward.

Looking at the Python code, there are plenty of `if a is not None` (same for ubound and xbound). This isn't different from what I propose.

The reasons why I changed it are:
- I realized none of the implemented constraints actually use both (do you have any in mind ?)
- most importantly, in Python, it is very easy to return several values whose type can change dynamically (e.g. F and g). In C++, this is not very easy to do and unexpected. I do not saying the API I implemented is the best. Now, if you would like to change it, we have to agree on the API. Before we do that, please have a look at the first draft implementation of the qpOASES wrapper in implemented just now, on my branch *solver*, in file *src/solver/qpOASES-wrapper.cpp*.

hungpham2511: > > I would prefer something simple and straightforward.
> 
> Looking at the Python code, there are plenty of `if a is not None` (same for ubound and xbound). This isn't different from what I propose.

There are a lot of if/else precisely because there is a single interface. I had in mind something similar for the C++ code actually (checking for empty matrices). 



> * most importantly, in Python, it is very easy to return several values whose type can change dynamically (e.g. F and g). In C++, this is not very easy to do and unexpected. I do not saying the API I implemented is the best. Now, if you would like to change it, we have to agree on the API. Before we do that, please have a look at the first draft implementation of the qpOASES wrapper in implemented just now, on my branch _solver_, in file _src/solver/qpOASES-wrapper.cpp_.

Maybe I am missing something, but I am not sure what is the problem with having a single interface with both (a,b,c,F,g) form and (ub,xb). Perhaps you can kindly point that out? 

For example, I can just make `ub, xb` have shape (N,0) instead of None and concatenate matrices normally.

jmirabel: This forces the allocation of several useless std::vector of N empty matrices and N might be a large number.

I will change the code and have two options to remove the BoxConstraint class:
1. add LinearConstraint::computeBounds,
2. change the API of LinearConstraint::computeParams. If I do this, I think it is better if the reference arguments are allocated by LinearConstraint::computeParams. As a bonus, this would avoid the issue I mentioned here https://github.com/hungpham2511/toppra/pull/83#discussion_r408898612.

hungpham2511: > This forces the allocation of several useless std::vector of N empty matrices and N might be a large number.
> 
> I will change the code and have two options to remove the BoxConstraint class:
> 
> 1. add LinearConstraint::computeBounds,

Right, so basically merge the methods in BoxConstraint back to LinearConstraint, is that right? I think we will also need a method to check if the (a,b,c,F,g) form exists.

> 2. change the API of LinearConstraint::computeParams. If I do this, I think it is better if the reference arguments are allocated by LinearConstraint::computeParams. As a bonus, this would avoid the issue I mentioned here [#83 (comment)](https://github.com/hungpham2511/toppra/pull/83#discussion_r408898612).

That's right. In fact, we can have a method that computes the required number of rows and columns, which can be used by the solver wrapper to allocate matrices if that is necessary. 

Btw, I think people usually use N ~ 100 to 1000. I know one guy even use N ~ 20.

jmirabel: I update the code accordingly.

hungpham2511:
Thanks! I will merge the PR.

PR #85, opened 2020-04-16
Feature: GeometricPath and PiecewisePolyPath (WIP)

hungpham2511:
Related to #82 

Work in progress at the moment. Will flesh out more details in the coming days.

### DIFF
--- a/cpp/src/toppra/geometric_path.hpp
+++ b/cpp/src/toppra/geometric_path.hpp
@@ -17,9 +17,11 @@ class PiecewisePolyPath : public GeometricPath {
 public:
   PiecewisePolyPath(const Matrices &, const std::vector<value_type> &);
   Vector eval(value_type, int degree = 0);
+  Vectors eval(std::vector<value_type>, int degree = 0);
 
+  size_t findSegmentIndex(value_type pos) const;
 private:
-  Matrices m_coefficients;
+  Matrices m_coefficients, m_coefficients_1, m_coefficients_2;
   std::vector<value_type> m_breakpoints;
   int m_dof, m_degree;
 };
--- a/cpp/src/toppra/geometric_path.hpp
+++ b/cpp/src/toppra/geometric_path.hpp
@@ -10,27 +10,26 @@
 namespace toppra {
 class GeometricPath {
 public:
-  GeometricPath(){};
-
+  virtual ~GeometricPath(){};
   /**
    * /brief Evaluate the path at given position.
    */
-  virtual Vector eval(value_type, int order = 0);
+  virtual Vector eval(value_type, int order = 0)=0;
 
   /**
    * /brief Evaluate the path at given positions (vector).
    */
-  virtual Vectors eval(std::vector<value_type>, int order = 0);
+  virtual Vectors eval(std::vector<value_type>, int order = 0)=0;
 
   /**
    * Return the degrees-of-freedom of the path.
    */
-  virtual int dof();
+  virtual int dof()=0;
 
   /**
    * Return the starting and ending path positions.
    */
-  virtual Vector pathInterval();
+  virtual Vector pathInterval()=0;
 };
 
 
@@ -59,6 +58,11 @@ public:
    * /brief Evaluate the path at given positions (vector).
    */
   Vectors eval(std::vector<value_type>, int order = 0);
+
+  /**
+   * /brief Evaluate the path at given positions (vector).
+   */
+  Vectors eval(Vector, int order = 0);
 
   /**
    * Return the degrees-of-freedom of the path.
--- a/cpp/src/toppra/geometric_path.hpp
+++ b/cpp/src/toppra/geometric_path.hpp
@@ -11,21 +11,54 @@ namespace toppra {
 class GeometricPath {};
 
 /**
- * Piecewise polynomial geometric path.
+ * \brief Piecewise polynomial geometric path.
+ *
+ * A simple implemetation of a piecewise polynoamial geometric path.
+ *
  */
 class PiecewisePolyPath : public GeometricPath {
 public:
+  /**
+   * Consructor.
+   *
+   * @param coefficients Polynoamial coefficients.
+   * @param breakpoints Vector of breakpoints.
+   */
   PiecewisePolyPath(const Matrices &, const std::vector<value_type> &);
-  Vector eval(value_type, int degree = 0);
-  Vectors eval(std::vector<value_type>, int degree = 0);
 
-  size_t findSegmentIndex(value_type pos) const;
+  /**
+   * /brief Evaluate the path at given position.
+   */
+  Vector eval(value_type, int order = 0);
+
+  /**
+   * /brief Evaluate the path at given positions (vector).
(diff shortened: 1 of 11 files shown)


jmirabel:
I would like to be able to use toppra even when the configuration space is not a vector space. For class GeometricPath, this only means that the dimension of the configuration space (so for configs) and of the tangent space (so for velocity and acceleration) are different.

hungpham2511:
That's an interesting use case. I am happy to happy toppra working on it as well.

Do you have anything in mind for the abstract interface of the geometric path already?

jmirabel:
For my use case, I will have to plug GeometricPath and this class https://gepettoweb.laas.fr/hpp/hpp-core/doxygen-html/a00064.html
The two points I have in mind are:
- Lie groups (what I already mentioned) (see outputSize and outputDerivativeSize functions)
- the fact that the interval of definition of a path is [T1,T2] (but the plugging class can handle that easily).

hungpham2511:
> For my use case, I will have to plug GeometricPath and this class https://gepettoweb.laas.fr/hpp/hpp-core/doxygen-html/a00064.html
> The two points I have in mind are:
> 
> * Lie groups (what I already mentioned) (see outputSize and outputDerivativeSize functions)
> * the fact that the interval of definition of a path is [T1,T2] (but the plugging class can handle that easily).

Can you recommend some readings on Lie Groups that is relevant to the implementation? I would like to understand better the theory to make an informed design choice.

Another question I have regarding Lie Group interpolation. Suppose we want to find time parametrization for a path in the configuration space, but with constraints on the rotation of some link -- Lie derivatives constraints -- what kind of design is required? 
- Probably some conversion class that transform the configuration-space path to a Lie group path? 
- Are we then able to reuse the Velocity/Acceleration constraint classes?

dinhhuy2109:
Not sure if TOPPRA has the same structure as TOPP,  but hopefully, you can found your answer here

https://github.com/dinhhuy2109/TOPP-SO3/blob/master/toppso3/Utils.py

I define Lietraj as (trans + rot) traj, since in SE(3) there is in general no bi-invariant interpolations.

For the constraints, I can reuse the same classes

You may have a look at https://github.com/dinhhuy2109/TOPP-SO3/blob/master/toppso3/lie.py 
and Park Ravani interpolation paper, my paper on TOPPSO3 for the literature.

hungpham2511:
> Not sure if TOPPRA has the same structure as TOPP, but hopefully, you can found your answer here
> 
> https://github.com/dinhhuy2109/TOPP-SO3/blob/master/toppso3/Utils.py
> 
> I define Lietraj as (trans + rot) traj, since in SE(3) there is in general no bi-invariant interpolations.
> 
> For the constraints, I can reuse the same classes
> 
> You may have a look at https://github.com/dinhhuy2109/TOPP-SO3/blob/master/toppso3/lie.py
> and Park Ravani interpolation paper, my paper on TOPPSO3 for the literature.

Awesome. Many thanks!

jmirabel:
For TOPP-RA algothim itself:
- with the collocation scheme, I think there isn't much to do.
- with the interpolation scheme, I didn't check the math yet, I have read the appendix of the article.

Then, I think the PiecewisePolyPath can be limited to R^n (at least until we achieve a first working version).
I don't know exactly how people use this library. From my side, I will have to write my own class inheriting from GeometricPath. So the only hard requirement for me is that GeometricPath supports the fact that the configuration and the velocity have different size and it doesn't have to implement interpolation in Lie groups. I think this is a reasonable first step.

cpp/src/toppra/geometric_path.hpp
```diff
@@ -8,7 +8,31 @@
 #include <vector>
 
 namespace toppra {
-class GeometricPath {};
+class GeometricPath {
+public:
+  GeometricPath(){};
```
jmirabel, line 13: You also need a virtual destructor. I always forget it.

hungpham2511:
I am done with the ppoly class. However I am having this compilation error when adding the virtual methods on the abstract base class GeometricPath. Any pointers would be great @jmirabel I am not very proficient in c++.

```
../src/libtoppra.so: undefined reference to `typeinfo for toppra::GeometricPath'
../src/libtoppra.so: undefined reference to `vtable for toppra::GeometricPath'
collect2: error: ld returned 1 exit status
tests/CMakeFiles/all_tests.dir/build.make:123: recipe for target 'tests/all_tests' failed
make[2]: *** [tests/all_tests] Error 1
CMakeFiles/Makefile2:349: recipe for target 'tests/CMakeFiles/all_tests.dir/all' failed
make[1]: *** [tests/CMakeFiles/all_tests.dir/all] Error 2
Makefile:129: recipe for target 'all' failed
make: *** [all] Error 2
```

By the way, feel free to review the PR. I propose merging this first, then working out #86

cpp/src/toppra/geometric_path.hpp
```diff
@@ -8,7 +8,31 @@
...
+
+  /**
+   * /brief Evaluate the path at given position.
+   */
+  virtual Vector eval(value_type, int order = 0);
+
+  /**
+   * /brief Evaluate the path at given positions (vector).
+   */
+  virtual Vectors eval(std::vector<value_type>, int order = 0);
```
jmirabel, line 23: Any reason not to use `Vector` here ? We will have to synchronize as I used `Vector` as type for vector of times.

hungpham2511: Noted. Will change.

cpp/src/toppra/geometric_path.hpp
```diff
@@ -1,11 +1,89 @@
...
 namespace toppra {
-  class GeometricPath {
+class GeometricPath {
+public:
+  GeometricPath(){};
+
+  /**
+   * /brief Evaluate the path at given position.
+   */
+  virtual Vector eval(value_type, int order = 0);
```
jmirabel, line 18: You need to make this method pure virtual. Add `= 0` before `;`.

cpp/src/toppra/geometric_path.hpp
hungpham2511: Basically I use std::vector for both coefficients and breakpoints, so for the sake of consistency I use std::vector there. Not a very strong reason though.

jmirabel: Maybe here, we could provide a default implementation that calls repeatedly
```
eval(time[i], order);
```

jmirabel: > Basically I use std::vector for both coefficients and breakpoints, so for the sake of consistency I use std::vector there. Not a very strong reason though.

In some places, I use the Eigen interface, e.g. https://github.com/jmirabel/toppra/blob/34e2a3f222374ac43a93936edefcd030c2512fc2/cpp/src/toppra/solver.cpp#L15

cpp/src/toppra/geometric_path.hpp
jmirabel: Just in case: `virtual ~GeometricPath() {}`

PR #90, opened 2020-04-22
Feature: TOPPRA algorithm

hungpham2511:
This PR:
- implements the TOPPRA algorithm.
- adds tests for polynomial geometric path class.
- adds tests for TOPPRA algorithm, comparing output with the Python codebase.

Comments are welcomed.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -14,10 +14,22 @@ jobs:
     # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
     - uses: actions/checkout@v2
 
+    - name: Install deps
+      run: |
+        sudo apt install -y curl
+        echo "deb [arch=amd64] http://robotpkg.openrobots.org/packages/debian/pub $(lsb_release -sc) robotpkg" \
+          | sudo tee /etc/apt/sources.list.d/robotpkg.list
+        curl http://robotpkg.openrobots.org/packages/debian/robotpkg.key \
+          | sudo apt-key add -
+        sudo apt update
+        sudo apt install -y clang-tidy libeigen3-dev robotpkg-pinocchio robotpkg-qpoases
+
     - name: Lint
       run: |
-        sudo apt install -y clang-tidy
         cd $GITHUB_WORKSPACE/cpp
+        export LD_LIBRARY_PATH=/opt/openrobots/lib:${LD_LIBRARY_PATH}
+        export CMAKE_PREFIX_PATH=/opt/openrobots
+        mkdir build && cd build && cmake -DBUILD_WITH_PINOCCHIO=ON -DBUILD_WITH_qpOASES=ON -DCMAKE_EXPORT_COMPILE_COMMANDS=1 ..
         clang-tidy  src/**/*.cpp \
-        -checks=clang-analyzer-*,clang-analyzer-cplusplus*,cppcoreguidelines-*,performance-*,modernize-*,readability-,-cppcoreguidelines-pro-bounds-array-to-pointer-decay\
-        -warnings-as-errors=clang-analyzer-*,clang-analyzer-cplusplus*,cppcoreguidelines-*,modernize-*
+          -checks=clang-analyzer-*,clang-analyzer-cplusplus*,cppcoreguidelines-*,performance-*,modernize-*,readability-,-cppcoreguidelines-pro-bounds-array-to-pointer-decay\
+          -warnings-as-errors=clang-analyzer-*,clang-analyzer-cplusplus*,cppcoreguidelines-*,modernize-*
--- a/cpp/CMakeLists.txt
+++ b/cpp/CMakeLists.txt
@@ -7,7 +7,7 @@ set(CMAKE_CXX_STANDARD_REQUIRED True)
 option(BUILD_WITH_PINOCCHIO "Compile with Pinocchio library" OFF)
 # Solvers
 option(BUILD_WITH_qpOASES "Compile the wrapper for qpOASES" OFF)
-option(TOPPRA_DEBUG_ON "Set debug mode to on." OFF)
+option(TOPPRA_DEBUG_ON "Set logging mode to on." OFF)
 
 find_package (Threads)
 
--- a/cpp/CMakeLists.txt
+++ b/cpp/CMakeLists.txt
@@ -7,6 +7,7 @@ set(CMAKE_CXX_STANDARD_REQUIRED True)
 option(BUILD_WITH_PINOCCHIO "Compile with Pinocchio library" OFF)
 # Solvers
 option(BUILD_WITH_qpOASES "Compile the wrapper for qpOASES" OFF)
+option(TOPPRA_DEBUG_ON "Set debug mode to on." OFF)
 
 find_package (Threads)
 
--- a/cpp/src/toppra/algorithm.cpp
+++ b/cpp/src/toppra/algorithm.cpp
@@ -1,3 +1,4 @@
+#include "toppra/toppra.hpp"
 #include <cstddef>
 #include <iostream>
 #include <toppra/algorithm.hpp>
@@ -26,27 +27,30 @@ ReturnCode PathParametrizationAlgorithm::computePathParametrization(double vel_s
 
 ReturnCode PathParametrizationAlgorithm::computeControllableSets(
     const Bound &vel_ends) {
+
+  LOG_DEBUG("computeControllableSets");
   ReturnCode ret = ReturnCode::OK;
   bool solver_ret;
   Vector g_upper{2}, g_lower{2}, solution;
   g_upper << 1e-9, -1;
   g_lower << -1e-9, 1;
-  m_data.controllable_sets(m_N, 0) = vel_ends(0);
-  m_data.controllable_sets(m_N, 1) = vel_ends(1);
+  m_data.controllable_sets(m_N, 0) = pow(vel_ends(0), 2);
+  m_data.controllable_sets(m_N, 1) = pow(vel_ends(1), 2);
 
   Matrix H;
   Bound x, x_next;
   x << 0, 100;
   x_next << 0, 1;
-  for (std::size_t i = m_N - 1; i >= 0; i--) {
-    // x_next << controllable_sets(i + 1, 0), controllable_sets(i + 1, 1);
+  for (std::size_t i = m_N - 1; i != (std::size_t) - 1; i--) {
+    LOG_DEBUG(i << ", " << m_N) ;
+    x_next << m_data.controllable_sets(i + 1, 0), m_data.controllable_sets(i + 1, 1);
     solver_ret =
         m_solver->solveStagewiseOptim(m_N - 1, H, g_upper, x, x_next, solution);
-    // std::cout << "up: " << solution << std::endl;
+    std::cout << "up: " << solution << std::endl;
 
     if (!solver_ret) {
       ret = ReturnCode::ERR_FAIL_CONTROLLABLE;
-      std::cout << "Fail: controllable, upper problem, idx: " << i << std::endl;
+      LOG_DEBUG("Fail: controllable, upper problem, idx: " << i);
       break;
     }
 
@@ -54,11 +58,12 @@ ReturnCode PathParametrizationAlgorithm::computeControllableSets(
 
     solver_ret =
         m_solver->solveStagewiseOptim(m_N - 1, H, g_lower, x, x_next, solution);
-    // std::cout << "down: " << solution << std::endl;
+
+    LOG_DEBUG("down: " << solution);
 
     if (!solver_ret) {
       ret = ReturnCode::ERR_FAIL_CONTROLLABLE;
-      std::cout << "Fail: controllable, lower problem, idx: " << i << std::endl;
+      LOG_DEBUG("Fail: controllable, lower problem, idx: " << i);
       break;
     }
 
--- a/cpp/src/toppra/algorithm.cpp
+++ b/cpp/src/toppra/algorithm.cpp
@@ -8,15 +8,15 @@ PathParametrizationAlgorithm::PathParametrizationAlgorithm(
     const LinearConstraintPtrs &constraints, const GeometricPath &path)
     : m_constraints(constraints), m_path(path){};
 
-ReturnCode PathParametrizationAlgorithm::computePathParametrization(
-    double vel_start, double vel_end) {
+ReturnCode PathParametrizationAlgorithm::computePathParametrization(double vel_start,
+                                                                    double vel_end) {
   ReturnCode ret;
   initialize();
   m_solver->setupSolver();
   Bound vel_ends;
   vel_ends << vel_end, vel_end;
   ret = computeControllableSets(vel_ends);
-  if ((int) ret > 0) {
+  if ((int)ret > 0) {
     return ret;
   }
   ret = computeForwardPass(vel_start);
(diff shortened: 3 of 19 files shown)


jmirabel:
Did you forgot to commit some files ?

cpp/src/toppra/solver.hpp
```diff
@@ -114,6 +114,8 @@ class Solver {
     Vector m_deltas;
 
 }; // class Solver
+
+typedef std::shared_ptr<Solver> SolverPtr;
```
jmirabel, line 118: This should go into `toppra/toppra.hpp`.

cpp/src/toppra/algorithm.hpp
```diff
@@ -1,6 +1,41 @@
...
+                               const GeometricPath &path)
+      : m_constraints(constraints), m_path(path){};
+
+  /** \brief Set the level of discretization used by the solver.
+   *
+   * If is zero, will attempt to detect automatically the most suitable grid.
+   */
+  void setN(int N) { m_N = N; };
+
+  ParametrizationData getParameterizationData();
```
jmirabel, line 28: The function is declared but never defined.

cpp/src/toppra/algorithm/toppra.cpp
```diff
@@ -0,0 +1,25 @@
...
+
+TOPPRA::TOPPRA(const LinearConstraintPtrs &constraints,
+               const GeometricPath &path)
+    : PathParametrizationAlgorithm{constraints, path} {};
+
+int TOPPRA::computePathParametrization(Vector &path_parametrization) {
+  m_solver = std::make_shared<solver::qpOASESWrapper>(
+      m_constraints, m_path,
+      Vector::LinSpaced(m_N, m_path.pathInterval()(0),
+                        m_path.pathInterval()(1)));
```
hungpham2511, line 19: Interestingly this dummy code does not run. I have this error. Have not really look into it though, but any idea?

```
unknown file: Failure
C++ exception with description "Invalid times." thrown in the test body.
```

cpp/src/toppra/algorithm.hpp
hungpham2511: Yes, it's pretty much work in progress now. This function only makes senses after the full algorithm is implemented.

cpp/src/toppra/algorithm/toppra.cpp
jmirabel: No. It doesn't ring a bell. The exception message is not very informative. Maybe it could be changed.

hungpham2511:
As the current stage, the code can compute the controllable sets, which is approximately half of the TOPPRA algorithm. The other half is much easier to implement now that everything is in place. 

However, the algorithm can't solve a simple instance that I throw at it, which means there are bugs in the current implementation somewhere. I propose to carry on with the implementation so that we can agree on the code, we get this PR merge before bug hunting. What do you think? @jmirabel

jmirabel:
> I propose to carry on with the implementation so that we can agree on the code, we get this PR merge before bug hunting.

Go ahead.

hungpham2511:
Preliminary inspection suggests that the coefficients passed to qpoases are incorrect, which is not entirely surprising. I will start to look into the issues.

Meanwhile, feel free to review and comment on the API. Thanks!

cpp/src/toppra/algorithm/toppra.cpp
jmirabel: #94 should fix this issue.

hungpham2511: I apply #94 on top of this PR and was able to run the full algorithm without any issue. I am comparing the output parametrization with the Python side and see if they are the same. If everything goes well we should have working code before very soon.

.github/workflows/main.yml
```diff
@@ -14,10 +14,22 @@ jobs:
...
+        sudo apt update
+        sudo apt install -y clang-tidy libeigen3-dev robotpkg-pinocchio robotpkg-qpoases
+
     - name: Lint
       run: |
-        sudo apt install -y clang-tidy
         cd $GITHUB_WORKSPACE/cpp
+        export LD_LIBRARY_PATH=/opt/openrobots/lib:${LD_LIBRARY_PATH}
+        export CMAKE_PREFIX_PATH=/opt/openrobots
+        mkdir build && cd build && cmake -DBUILD_WITH_PINOCCHIO=ON -DBUILD_WITH_qpOASES=ON -DCMAKE_EXPORT_COMPILE_COMMANDS=1 ..
```
jmirabel, line 32: Is there any reason to use two CI ? I think one of them could do everything.

hungpham2511: I was just trying out the new Github Action to see if there is anything interesting. Nothing interesting so far. The Github server seems to be slower quite a bit as well. At this rate we might as well use CircleCI for linting for the C++ code.

jmirabel: I think one CI is better on many aspects (less maintenance, less energy...).

hungpham2511: Sure, totally agree. I have the bad habit of having to try new things out... This time it quite clear that CircleCI is fully capable of doing whatever we need it to do.

Luckily it looks like they (CIs) are basically copying each other, so moving from on to another is not very difficult.

hungpham2511:
Look like we have a working TOPRA algorithm now. The C++ test pass. Can you help to review and accept the PR? @jmirabel

cpp/src/toppra/algorithm.cpp
```diff
@@ -0,0 +1,80 @@
...
+ReturnCode PathParametrizationAlgorithm::computeControllableSets(
+    const Bound &vel_ends) {
+  TOPPRA_LOG_DEBUG("computeControllableSets");
+  ReturnCode ret = ReturnCode::OK;
+  bool solver_ret;
+  Vector g_upper{2}, g_lower{2}, solution;
+  g_upper << 1e-9, -1;
+  g_lower << -1e-9, 1;
+  m_data.controllable_sets(m_N, 0) = pow(vel_ends(0), 2);
+  m_data.controllable_sets(m_N, 1) = pow(vel_ends(1), 2);
```
jmirabel, line 37:
```suggestion
  m_data.controllable_sets.row(m_N) = vel_ends.cwiseAbs2();
```

cpp/src/toppra/algorithm.cpp
```diff
@@ -0,0 +1,80 @@
...
+
+  Matrix H;
+  Bound x, x_next;
+  x << 0, 100;
+  x_next << 0, 1;
+  for (std::size_t i = m_N - 1; i != (std::size_t)-1; i--) {
+    TOPPRA_LOG_DEBUG(i << ", " << m_N);
+    x_next << m_data.controllable_sets(i + 1, 0), m_data.controllable_sets(i + 1, 1);
```
jmirabel, line 45:
```suggestion
    x_next = m_data.controllable_sets.row(i + 1);
```

cpp/src/toppra/algorithm.cpp
```diff
@@ -0,0 +1,80 @@
...
+};
+
+ReturnCode PathParametrizationAlgorithm::computeControllableSets(
+    const Bound &vel_ends) {
+  TOPPRA_LOG_DEBUG("computeControllableSets");
+  ReturnCode ret = ReturnCode::OK;
+  bool solver_ret;
+  Vector g_upper{2}, g_lower{2}, solution;
+  g_upper << 1e-9, -1;
+  g_lower << -1e-9, 1;
```
jmirabel, line 35: what is this 1e-9 for ?

cpp/CMakeLists.txt
```diff
@@ -7,6 +7,7 @@ set(CMAKE_CXX_STANDARD_REQUIRED True)
 option(BUILD_WITH_PINOCCHIO "Compile with Pinocchio library" OFF)
 # Solvers
 option(BUILD_WITH_qpOASES "Compile the wrapper for qpOASES" OFF)
+option(TOPPRA_DEBUG_ON "Set debug mode to on." OFF)
```
jmirabel, line 10: If this option turns on the log, I would suggest do
```suggestion
option(TOPPRA_DEBUG_ON "Set logging mode to on." OFF)
```

cpp/src/toppra/algorithm.hpp
```diff
@@ -1,6 +1,106 @@
...
+  /** \brief Compute the forward pass.
+   *
+   * Derived class should provide a suitable forward pass function,
+   * depending on the desired objective.
+   */
+  virtual ReturnCode computeForwardPass(double vel_start) = 0;
+
+  /** Compute the sets of feasible squared velocities.
+   */
+  ReturnCode computeFeasibleSets() { throw std::runtime_error("Not implemented"); };
```
jmirabel, line 85: Is it possible to write a PathParameterizationAlgorithm with implementing this ? If no, make it pure virtual, otherwise forget about this comment.

cpp/src/toppra/algorithm/toppra.cpp
```diff
@@ -0,0 +1,41 @@
...
+  bool solver_ret;
+  Vector g_upper{2}, solution;
+  Matrix H;
+  auto deltas = m_solver->deltas();
+  // std::cout << deltas << std::endl;
+  Bound x, x_next;
+  m_data.parametrization(0) = vel_start;
+  for (std::size_t i = 0; i < m_N; i++) {
+    g_upper << -2 * deltas(i), -1;
+    x << m_data.parametrization(i), m_data.parametrization(i);
```
jmirabel, line 24:
```suggestion
    x.setConstant(m_data.parametrization(i));
```

cpp/src/toppra/algorithm/toppra.cpp
```diff
@@ -0,0 +1,41 @@
...
+    x_next << m_data.controllable_sets(i + 1, 0), m_data.controllable_sets(i + 1, 1);
```
jmirabel, line 25:
```suggestion
    x_next = m_data.controllable_sets.col(i + 1);
```

cpp/src/toppra/toppra.hpp
```diff
@@ -2,11 +2,19 @@
...
 
 #include <Eigen/Core>
 #include <Eigen/StdVector>
 #include <vector>
 
+#ifdef TOPPRA_DEBUG_ON
+#define TOPPRA_LOG_DEBUG(X) std::cout << "[DEBUG]: " <<  X << std::endl
+#else
+#define TOPPRA_LOG_DEBUG(X) if (true) {} else {std::cout << "[DEBUG]: " <<  X << std::endl;}
+#endif
```
jmirabel, line 15: Here you can choose between
```cpp
#ifdef TOPPRA_DEBUG_ON
#define TOPPRA_LOG_DEBUG(X) std::cout << "[DEBUG]: " <<  X << std::endl
#else
#define TOPPRA_LOG_DEBUG(X) ((void)0) // this is just to avoid compiler warnings.
#endif
```
or
```cpp
#ifdef TOPPRA_DEBUG_ON
#define TOPPRA_LOG_DEBUG std::cout << "[DEBUG]: "
#else
#define TOPPRA_LOG_DEBUG if (true) {} else std::cout << "[DEBUG]: "
#endif
```

The two versions are used as
```cpp
TOPPRA_LOG_DEBUG("my message" << my_variable); // version 1
TOPPRA_LOG_DEBUG << "my message" << my_variable << std::endl; // version 2
```

cpp/tests/test_algorithm.cpp
```diff
@@ -0,0 +1,125 @@
...
+// x_eval = [0, 0.5, 1., 1.1, 2.5]
+// print("Eval for x_eval = {:}\npath(x_eval)=\n{}\npath(x_eval, 1)=\n{}\npath(x_eval,
+// 2)=\n{}".format(
+//     x_eval, path(x_eval), path(x_eval, 1), path(x_eval, 2)))
+//
+// pc_vel = ta.constraint.JointVelocityConstraint([1.0, 1.0])
+// pc_acc = ta.constraint.JointAccelerationConstraint([0.2, 0.2])
+//
+// instance = ta.algorithm.TOPPRA([pc_vel, pc_acc], path, gridpoints=np.linspace(0, 3,
+// 51)) sdds, sds, _ = instance.compute_parameterization(0, 0)
```
jmirabel, line 45: Could you put the python code between `/* */` instead to make it is copy-pastable ?

cpp/tests/test_poly_geometric_path.cpp
```diff
@@ -1,6 +1,8 @@
 #include "gtest/gtest.h"
 #include <Eigen/Dense>
 
+#include <Eigen/src/Core/util/Constants.h>
+#include <gmock/gmock-matchers.h>
```
jmirabel, line 5:
```suggestion
```

I think you don't need this two headers. With Eigen, you should never include stuff in `Eigen/src` yourself. I think the second is not used below.

cpp/tests/test_poly_geometric_path.cpp
```diff
@@ -135,3 +145,113 @@ TEST_F(BadInputs, ThrowIfWrongNumberOfBreakPoints) {
...
+//                 out += "{:f};\n".format(t)
+//             else:
+//                 out += "{:f}, ".format(t)
+//     return out
+// 
+// print(print_cpp_code(path))
+// print("breakpoints: {}".format([0, 1, 2, 3]))
+// x_eval = [0, 0.5, 1., 1.1, 2.5]
+// print("Eval for x_eval = {:}\npath(x_eval)=\n{}\npath(x_eval, 1)=\n{}\npath(x_eval, 2)=\n{}".format(
+//     x_eval, path(x_eval), path(x_eval, 1), path(x_eval, 2)))
```
jmirabel, line 170: Python code between `/* */` makes it copy-pastable.

cpp/src/toppra/algorithm.cpp
hungpham2511: Cool! I don't know this is possible.

cpp/src/toppra/algorithm.cpp
hungpham2511: It's the value used in the Python implementation. I vaguely remember that it's used to break some tie, but couldn't remember exactly why.

cpp/src/toppra/algorithm.hpp
hungpham2511: It's possible. Will do.

cpp/src/toppra/algorithm/toppra.cpp
hungpham2511: Shouldn't it be row?

cpp/src/toppra/toppra.hpp
hungpham2511: Thanks for the suggestion. I will go with the first one since there are quite a few debug messages in this form already.

cpp/src/toppra/algorithm/toppra.cpp
jmirabel: You are right.

cpp/src/toppra/algorithm/toppra.cpp
```diff
@@ -29,7 +28,7 @@ ReturnCode TOPPRA::computeForwardPass(double vel_start) {
       TOPPRA_LOG_DEBUG("Fail: forward pass, idx: " << i);
       break;
     }
-    // TODO: This can be optimized further by solving a 1D problem instead of 2D
+    // \todo This can be optimized further by solving a 1D problem instead of 2D
```
jmirabel, line 31: If you want doxygen to read this, you must add a third `/` in the beginning.

jmirabel:
```suggestion
    /// \todo This can be optimized further by solving a 1D problem instead of 2D
```

PR #94, opened 2020-04-24
Fix bug in qpOASESWrapper.

jmirabel:


cpp/src/toppra/solver/qpOASES-wrapper.cpp
```diff
@@ -63,13 +63,11 @@ bool qpOASESWrapper::solveStagewiseOptim(std::size_t i,
...
     m_A.row(0) << -2 * delta, -1;
     m_hA[0] = - xNext[0];
     m_lA[0] = - m_boundary;
 
-    // TODO self._A[1] access 1-th row ?
-    m_A.row(1).setZero();
+    m_A.row(1) << 2 * delta, 1;
     m_hA[1] = xNext[1];
     m_lA[1] = -m_boundary;
   }
```
jmirabel, line 73: Here, when `i == N`, then the 2 first rows of `A, lA, hA` are left uninitialized. This code is mostly a copy-paste of the Python code and the same issue (if it is an issue) appears in the Python code. Is this expected ?

hungpham2511: In the Python code, for i == N, the inequalities become trivially true. This is because A[:2, :] is zero and lA and hA are initialized to -INF and INF respectively.

hungpham2511:
Look like a straightforward bug fix. Please go ahead and merge whenever you are ready.

cpp/src/toppra/solver/qpOASES-wrapper.cpp
jmirabel: Alright, then I have to fix the C++ code.

jmirabel:
I added another bug fix.

cpp/tests/test_poly_geometric_path.cpp
```diff
@@ -40,7 +38,7 @@ class ConstructPiecewisePoly : public testing::Test {
 
 TEST_F(ConstructPiecewisePoly, OutputValueHasCorrectDOF) {
   toppra::Vector pos = path.eval_single(0.5, 0);
-  ASSERT_THAT(pos.rows(), testing::Eq(2));
+  ASSERT_EQ(pos.rows(), 2);
```
hungpham2511, line 41: I am pretty sure they are equivalent. Why is this a bug?

jmirabel: This was not the bug. I had a weird compilation error at some point, coming from somewhere in `gmock.h` and removing the `include <gmock.h>` fixed. I couldn't understand why, and as we don't really use it, this is satisfactory to me.
ASSERT_THAT coming from this file, I had to change it.

cpp/src/toppra/solver/qpOASES-wrapper.cpp
jmirabel: Should be fixed now.

jmirabel:
From my side, this is ready. I will merge when CI is happy.

jmirabel:
Oh, I do not have the rights to merge.

PR #97, opened 2020-04-25
Add solver based on GLPK.

jmirabel:
and add unit test that checks consistency between qpOASESWrapper and GLPKWrapper.


cpp/src/toppra/solver/glpk-wrapper.cpp
```diff
@@ -0,0 +1,138 @@
+#include <toppra/solver/glpk-wrapper.hpp>
+
+#include <glpk.h>
+
+#include <iostream>
+
+namespace toppra {
+namespace solver {
+
+constexpr value_type infty = std::numeric_limits<value_type>::infinity();
```
hungpham2511, line 10: This should be in solver.hpp?

hungpham2511:
How do you plan to select the solver wrapper? Probably `toppra::algorithm::TOPPRA` should implement a `initialize_solver` method. Currently it's done here: 
https://github.com/hungpham2511/toppra/blob/c31982910a9b25c1eed750a4b6110f01428d0810/cpp/src/toppra/algorithm.cpp#L106

cpp/src/toppra/solver/glpk-wrapper.cpp
jmirabel: Yes, could be. I will do it.

jmirabel:
> How do you plan to select the solver wrapper?

This should go into another PR. We could implement it as follows:
- let PathParameterisationAlgorithm take any Solver derived class. This means the constructor of solvers should be empty and the initialization should go into an initialization function,
- add a function that *selects* a solver (chosen among the available ones),
- eventually, add a function that creates an instance of TOPPRA, setup with a *selected solver*.

This makes it possible to explicitly choose the solver if desired, while having a default choice when not caring about which solver is used. This is similar to what you did in Python, except that user will be able to plug their own solver without having to modifying this library.

jmirabel:
I add here a bit of CMake (see the README) and some minor changes.

jmirabel:
I wrote the code I mentioned above. It is on branch refactoring_solver of my fork. Should I add it here ?

cpp/src/toppra/constraint.cpp
```diff
@@ -42,6 +47,10 @@ void allocateLinearPart(std::size_t N, Eigen::Index k, Eigen::Index m,
   for (Vector& x : g) x.resize(k);
 }
 
+void checkSizes (std::size_t N, Eigen::Index k, Eigen::Index m,
```
hungpham2511, line 50: Why are there two definitions of this function?

cpp/src/toppra/constraint.cpp
```diff
@@ -20,8 +20,13 @@ void LinearConstraint::discretizationType (DiscretizationType type)
 }
 
 /// \internal
+
 /// \param k number of constraints
 /// \param m number of variables
+void allocateLinearPart(std::size_t N, Eigen::Index k, Eigen::Index m,
+    bool constantF, Vectors& a, Vectors& b, Vectors& c, Matrices& F, Vectors& g)
+  TOPPRA_NO_EXPORT;
```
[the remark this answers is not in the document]
jmirabel: The file is generated by CMake. It tells the compiler that the function is internal and does not have to be accessible from outside.

cpp/src/CMakeLists.txt
```diff
@@ -1,3 +1,5 @@
+include(GenerateExportHeader)
```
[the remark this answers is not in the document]
jmirabel: It is used to fill the visibility attribute of a symbol. See https://gcc.gnu.org/wiki/Visibility

cpp/src/toppra/constraint.cpp
jmirabel: There is one declaration and one definition. Visibility flag can only be put in the declaration.

cpp/README.md
```diff
@@ -17,3 +17,12 @@ make -j4
 ./tests/all_tests
 ```
 
+# Using TOPPRA in CMake-based project
```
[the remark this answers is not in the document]
jmirabel: very simple: `sudo apt install libglpk-dev`. I will add it.

cpp/src/toppra/constraint.cpp
jmirabel: You may have a look at https://www.tutorialspoint.com/What-is-the-difference-between-a-definition-and-a-declaration-in-Cplusplus

hungpham2511: I see. So instead of putting the declaration in the header file, it is placed directly to the source file. This is possible because no other classes make use of this function anyway. 

Is it valid to put TOPPRA_NO_EXPORT to private class methods in header files?

hungpham2511:
> I wrote the code I mentioned above. It is on branch refactoring_solver of my fork. Should I add it here ?

Let's open a separate PR, it will help make the git log cleaner.

cpp/src/toppra/constraint.cpp
jmirabel: Yes, it is valid. I don't know whether it is useful because I don't what is the default for private class method.

The export stuff is generally required when supporting several compilers because all the compilers do not use the same default visibility. VC++ hides symbols by default while gcc makes them visible by default...

hungpham2511:
Hi @jmirabel, please merge when you are happy with the PR. I have added you as a collaborator. Many thanks for the contributions. 

I think the code base is functional now. What it needs is some "exposure" to find out bugs as well as use cases. Do you have anything in mind already? Please let me know if there is anything I can help with.

jmirabel:
> Do you have anything in mind already?

I started to use it in HPP but I had to switch to another task. I will come back to this next week.

jmirabel:
For reference, related to #82

PR #98, opened 2020-04-28
Eliminate use of deprecated method

mttamtam:
Since `SplineInterpolator`'s `eval`, `evald`, `evaldd` methods are deprecated, this PR replaces them to `__call__` method.
- This change reduces the number of warnings in pytest.
- New interpolators don't need to equip with `eval`, `evald`, `evaldd` methods.


hungpham2511:
Thanks for the work! Appreciate it.

PR #99, opened 2020-04-29
Update API to enable user to choose a solver.

jmirabel:


jmirabel:
Follows discussion in #97 
Related to #82

cpp/src/toppra/algorithm.cpp
```diff
@@ -1,14 +1,27 @@
...
 
 PathParametrizationAlgorithm::PathParametrizationAlgorithm(
-    LinearConstraintPtrs constraints, const GeometricPath &path)
-    : m_constraints(std::move(constraints)), m_path(path){};
+    LinearConstraintPtrs constraints, const GeometricPathPtr &path)
+    : m_constraints(std::move(constraints)), m_path(path),
+#ifdef BUILD_WITH_qpOASES
+    m_solver (std::make_shared<solver::qpOASESWrapper>())
+#elif BUILD_WITH_GLPK
+    m_solver (std::make_shared<solver::qpOASESWrapper>())
```
hungpham2511, line 20: This looks like a Typo.

Btw, I think we should delay solver initialization in the derived class to avoid explicit dependencies in the base PathParametrizationAlgorithm class.

jmirabel: Good catch !

PR #107, opened 2020-05-07
[python] Feature: Parametrization outputs more information 

### LINKED ISSUE #80: Users should be informed when there is failure
The parametrization algorithm might fail when:
1. There is actually no feasible solution;
2. The output trajectory does not satisfy input constraints.
3. ??? 

Users should be notified when any of these errors occur.

hungpham2511:
Fixes #80 

Changes in this PRs:
- Modify the behavior of parametrization algorithms to report more information when they fail

Checklists:


jmirabel:
Hello @hungpham2511,

do you think it is possible to prefix the commit / PR message with [C++] or [Python] so that we know on what this applies ?
I think it will help clarifying the logs and, personally, I don't follow closely the Python part while I do for the C++ part so it would help me filtering too.

hungpham2511:
Sure. Thanks for the suggestion.

PR #110, opened 2020-05-23
[cpp,python]Feature: Improve PiecewisePolyPath

hungpham2511:
Changes in this PRs:
- Provide bindings with pybind for PiecewisePolyPath
- Serialization/desrialization with msgpack;
- Construct a PiecewisePolyPath as a Hermite Spline

Checklists:


hungpham2511:
Hi @jmirabel I made a few changes that might be breaking (e.g. add a few CMake flags), but can be easily turn off. If you have any comments please let me know.

cpp/src/toppra/geometric_path/piecewise_poly_path.hpp
```diff
@@ -0,0 +1,70 @@
...
+
+  /**
+   * Return the starting and ending path positions.
+   */
+  Bound pathInterval() const;
+  void serialize(std::ostream &O) const override;
+  void deserialize(std::istream &I) override;
+
+  void constructHermite(const Vectors &positions, const Vectors &velocities,
+                        const std::vector<value_type> times);
```
jmirabel, line 55: A minor comment: API-wise, I think it is better to make this function static and return a newly allocated object.

jmirabel:
I see no problem with those changes. Are they really breaking the API or just adding new functions ?

cpp/CMakeLists.txt
```diff
@@ -14,6 +14,10 @@ option(BUILD_WITH_qpOASES "Compile the wrapper for qpOASES" OFF)
 option(BUILD_WITH_GLPK "Compile the wrapper for GLPK (GPL license)" OFF)
 # General options
 option(TOPPRA_DEBUG_ON "Set logging mode to on." OFF)
+option(PYTHON_BINDINGS "Build bindings for Python" ON)
```
hungpham2511, line 17: @jmirabel You will need to turn this off if there is no pybind installation.

jmirabel: Ok. Thanks for the notice.

PR #114, opened 2020-05-29
[cpp]Feature: Bindings for multiple components

### LINKED ISSUE #103: Python binding for the C++ API
The C++ API is, in every way, superior to the Python API when it comes to speed, low-level access to solvers. Further maintaining two APIs at the same time is time-consuming.

Thus, creating Python binding for the C++ API looks like a good addition to the library.

**Checklist**
- [x] Trajectory #110 ;
- [ ] ParametrizeConstAccel (generate output traj from a base traj and a parametrization);
- [ ] Constraints;
- [ ] ~Solvers~ (not necessary);
- [ ] Full algorithms.

hungpham2511:
Fixes #103 

Changes in this PRs:
- Provide most of the binding needed to run the C++ API from python.

Checklists:
- [ ] Update CHANGELOG.md with a single line describing this PR


hungpham2511:
@jmirabel I have 1 question regarding the usage of the joint torque constraint, in this code snippet that you wrote in the test, I suppose the kinematics is loaded in the second line. Questions:

- Is this a generic model that you use for testing only?
- How can I load a model, say from urdf? 

```
  JointTorque::Model model;
  pinocchio::buildModels::manipulator(model);
#if PINOCCHIO_VERSION_AT_MOST(2,4,1)
  // Work around bug solved by https://github.com/stack-of-tasks/pinocchio/pull/1155
  model.effortLimit.setConstant(10);
#endif
  Vector frictions (Vector::Constant(model.nv, 0.001));
  JointTorque constraint (model, frictions);

```

jmirabel:
I will provide a commit that binds class jointTorque::Pinocchio, and that uses pinocchio python bindings. You can remove your two last commits.

> * Is this a generic model that you use for testing only?

In which language ?

      pinocchio::buildModels::manipulator(model); // C++
      model = pinocchio.buildSampleModelManipulator()  # Python

This model sufficient for testing purposes.

> 
> * How can I load a model, say from urdf?

```python
import pinocchio
model = pinocchio.buildModelFromUrdf(urdf_filename)
```
If that suits you, package https://github.com/Gepetto/example-robot-data/ (binary package: `robotpkg-example-robot-data`) provides some robots. Then, in Python:
```
from example_robot_data import <loadfunction>
robot = <loadfunction>()
model = robot.model
```
where *loadfunction* is any of the function you find in https://github.com/Gepetto/example-robot-data/blob/master/python/example_robot_data/__init__.py

```c++
#include <pinocchio/parsers/urdf.hpp>

pinocchio::Model model;
pinocchio::urdf::buildModel(urdf_filename,model);
```

jmirabel:
I just see your last commit. We have to synchronize otherwise we will have plenty of git conflicts. I will let you merge this when ready and then I will provide the failing test on top of it.

I can take care of binding jointTorque::Pinocchio.

cpp/bindings/bindings.hpp
```diff
@@ -53,6 +62,73 @@ class PyPiecewisePolyPath {
...
+
+ private:
+  std::unique_ptr<toppra::algorithm::TOPPRA> m_problem;
+  LinearConstraintPtr m_tmp;
+};
+
+/**
+   Base Python wrapper for constraints.
+ */
+class PyLinearConstraint {
```
jmirabel, line 87: Why do you need to wrap LinearConstraint and TOPPRA ?
I think you can directly work with LinearConstraint, no ? In case the issue is memory storage, see https://pybind11.readthedocs.io/en/stable/advanced/smart_ptrs.html#std-shared-ptr

cpp/bindings/toppra_int.cpp
```diff
@@ -36,6 +39,83 @@ PYBIND11_MODULE(toppra_int, m) {
...
+      .def_property_readonly("nbVariables", &LinearConstraint::nbVariables)
+      .def_property_readonly("hasLinearInequalities",
+                             &LinearConstraint::hasLinearInequalities)
+      .def_property_readonly("hasUbounds", &LinearConstraint::hasUbounds)
+      .def_property_readonly("hasXbounds", &LinearConstraint::hasXbounds)
+      .def_property("discretizationType",
+                    (DiscretizationType(LinearConstraint::*)() const) &
+                        LinearConstraint::discretizationType,
+                    (void (LinearConstraint::*)(DiscretizationType)) &
+                        LinearConstraint::discretizationType);
```
jmirabel, line 73: You can share the bindings as follows:
```suggestion
  py::class_<LinearConstraint>(m, "_LinearConstraint")
      .def_property_readonly("nbConstraints", &LinearConstraint::nbConstraints)
      .def_property_readonly("nbVariables", &LinearConstraint::nbVariables)
      .def_property_readonly("hasLinearInequalities",
                             &LinearConstraint::hasLinearInequalities)
      .def_property_readonly("hasUbounds", &LinearConstraint::hasUbounds)
      .def_property_readonly("hasXbounds", &LinearConstraint::hasXbounds)
      .def_property("discretizationType",
                    (DiscretizationType(LinearConstraint::*)() const) &
                        LinearConstraint::discretizationType,
                    (void (LinearConstraint::*)(DiscretizationType)) &
                        LinearConstraint::discretizationType);

  py::class_<constraint::LinearJointVelocity, LinearConstraint>(m,
                                                                "LinearJointVelocity")
      .def(py::init<const Vector &, const Vector &>());
  py::class_<constraint::LinearJointAcceleration, LinearConstraint>(
      m, "LinearJointAcceleration")
      .def(py::init<const Vector &, const Vector &>());
```

cpp/bindings/bindings.cpp
```diff
@@ -25,45 +29,5 @@ nparr toNumpyArray(const toppra::Matrices& ret) {
   return x;
 }
 
-PyPiecewisePolyPath::PyPiecewisePolyPath(const Matrices& coefficients,
```
jmirabel, line 28: A few line above this line, at
          for (size_t k = 0; j < x.shape()[2]; j++) x.mutable_at(i, j, k) = ret[i](j, k);

you have a bug in the indices inside the *for* declaration.

cpp/bindings/py_geometric_path.cpp
```diff
@@ -0,0 +1,46 @@
...
+
+PyPiecewisePolyPath::PyPiecewisePolyPath(const PiecewisePolyPath& path)
+    : m_path(path){};
+
+Vector PyPiecewisePolyPath::eval_single(value_type x, int order) const {
+  return m_path.eval_single(x, order);
+};
+
+nparr PyPiecewisePolyPath::eval(const Vector& xs, int order) const {
+  auto ret = m_path.eval(xs, order);
```
jmirabel, line 17: Did you try to let pybind11 convert the vector of Eigen matrices by itself ?

Given https://pybind11.readthedocs.io/en/stable/advanced/cast/stl.html and https://pybind11.readthedocs.io/en/stable/advanced/cast/eigen.html, I think it should be able to do it.
If you did this to have the same type as in the Python implementation, I understand.

hungpham2511:
Sorry I went ahead and worked on a binding for the JointTorque::Pinocchio class as well.

This PR is still a WIP. I am changing the bindings of the constraints to uniformly using the PyLinearConstraint class. Construction at runtime must be done via the static constructor: 
```
  static PyLinearConstraint createJointTorquePinocchio(const std::string& urdf_name);

  static PyLinearConstraint createJointAcceleration(const Vector& lowerAlimit,
                                                    const Vector& upperAlimit);
```

I will remove the bindings of LinearJointVelocity and LinearJointAcceleration, as well as of LinearConstraint.

The reason underlying this design is because I *think* it's not possible to bind the JointTorque::Pinocchio class directly as the constructor requires the Model object.

cpp/bindings/py_geometric_path.cpp
hungpham2511: Pybind will convert Vector to np.ndarray, but the output type here is Vectors, which is converted to List[np.ndarray]

cpp/bindings/bindings.cpp
hungpham2511: Thanks!

cpp/bindings/bindings.hpp
hungpham2511: I tried working directly before this, it doesn't work. It's because the TOPPRA constructor looks for LinearConstraintPtrs. Yet, when I give the python bindings a list of different constraints, it is not cast properly to LinearConstraintPtrs.

hungpham2511:
> I will provide a commit that binds class jointTorque::Pinocchio, and that uses pinocchio python bindings. You can remove your two last commits.

I am happy to do so if Pinocchio works for Python3.7

cpp/bindings/py_geometric_path.cpp
jmirabel: For your info, you do not have to write a wrapper to achieve the same behavior. For function of `PiecewisePolyPath` that needs rework to be exposed in Python, you can write static function
```cpp
retType bind_function (PiecewisePolyPath& p, args...) {
res =  p.function (args);
// Do something with res and return it.
}
```
Then when binding
```cpp
py::class_<PiecewisePolyPath>(m, "PiecewisePolyPath")
.def ("dof", &PiecewisePolyPath::dof)
.def("function", bind_function)
;
```

You will have a lot less code. You can even write lambda functions (https://pybind11.readthedocs.io/en/stable/classes.html?highlight=lambda#binding-lambda-functions).

jmirabel:
> I am happy to do so if Pinocchio works for Python3.7

It does.

jmirabel:
> This PR is still a WIP. I am changing the bindings of the constraints to uniformly using the PyLinearConstraint class. Construction at runtime must be done via the static constructor:
> 
> ```
>   static PyLinearConstraint createJointTorquePinocchio(const std::string& urdf_name);
> 
>   static PyLinearConstraint createJointAcceleration(const Vector& lowerAlimit,
>                                                     const Vector& upperAlimit);
> ```
> 
> I will remove the bindings of LinearJointVelocity and LinearJointAcceleration, as well as of LinearConstraint.

Having a wrapper is always tedious. Are you sure you actually need this wrapper ? Cannot you achieve the Python API using pybind11 stuff only ?

jmirabel:
> The reason underlying this design is because I _think_ it's not possible to bind the JointTorque::Pinocchio class directly as the constructor requires the Model object.

It is possible. I did it and will put in on top of your work.

cpp/bindings/bindings.hpp
jmirabel: I think you must declare the LinearConstraint as
```
py::class_<LinearConstraint, LinearConstraintPtr> (...)
```

jmirabel: And you have to do the same for all the constraint classes.

hungpham2511:
I implement the wrappers precisely because I was not able to get the C++ API and pybind work together nicely.

Then I will revert my last two commits and you can apply your commit?

hungpham2511:
Done, I will wait until you have finished with the Pinocchio bindings.

In the lastest commit, the bindings can already be used for Velocity/Acceleration constraints. So I think after your addition we can start doing some evaluation.

hungpham2511:
> > I am happy to do so if Pinocchio works for Python3.7
> 
> It does.

Can we have another constructor that constructs new Pinocchio constraint given only the urdf name? Then I don't have to worry about installing Pinocchio's bindings.

PR #115, opened 2020-05-30
[cpp]Add an option to configure Python version when building the Python bindings

hungpham2511:
Changes in this PRs:
- Add an option to configure Python version when building the Python bindings
- Fix a numerical issue.

Checklists:


jmirabel:
From my side, this can be merged. No idea why CI is broken. I suggest to disable it until we understand why it does not run.

PR #116, opened 2020-05-30
[cpp] Update Python Bindings + fix numerical issues.

### LINKED ISSUE #112: [C++] numerical issue
Hi @hungpham2511 

I encounter a numerical issue at this line:
https://github.com/hungpham2511/toppra/blob/5ffa8e1b92c8151e837a2fd4253948860d9bbc38/cpp/src/toppra/solver/glpk-wrapper.cpp#L113

Function `intersection(a, b, c)` computes the intersection between interval a and interval b and stores it into c. At some point, I get `a = [0.11138810763690521, 0.11138810763690521 ]` and `b = [ 0, 0.11138810763690514 ]`. If you look carefully at the values, you will see that:
- the intersection is empty and
- the upper bound of `b` is very close to `a`.

I get this during the forward pass. In my scenario, I have only one linear velocity constraint (which gives `b`). When the fault happens, the `a` value is the `x` value passed to the solver (so the feasible set at the index `i`).

I can make a work around in this solver but I am wondering whether this should be addressed at a higher stage, like in the TOPPRA algorithm itself.

Any thoughts on this ?

jmirabel:
Fixes #112 (together with #111 and #115)

Changes in this PRs:
- Update the Python bindings.
- Fix numerical issues.
- Fix bug with interpolation discretisation scheme.
- Provide a script that check that C++ and Python implementation returns the same output.

Checklists:


jmirabel:
Should be merged before #114

cpp/bindings/CMakeLists.txt
```diff
@@ -11,11 +11,21 @@ if(${PYTHON_BINDINGS})
...
   set_target_properties(toppra_int PROPERTIES
     LIBRARY_OUTPUT_DIRECTORY ${TOPPRA_PYTHON_SOURCE_DIR}
     )
+  if(BUILD_WITH_PINOCCHIO)
+    target_compile_options(toppra_int PRIVATE -DBUILD_WITH_PINOCCHIO)
+    target_link_libraries(toppra_int PUBLIC pinocchio::pinocchio)
+    if(DEFINED PINOCCHIO_WITH_PYTHON_INTERFACE)
+      find_package(Boost REQUIRED COMPONENTS python3)
+      target_compile_options(toppra_int PRIVATE -DPINOCCHIO_WITH_PYTHON_INTERFACE)
+      target_link_libraries(toppra_int PUBLIC Boost::python3)
```
hungpham2511, line 27: This will require Boost right? I am a bit concern with adding boost, since it's a huge library.

hungpham2511: And some of my packages are running with an outdated version of it (1.5x) and I worry that things will not work properly because of this.

cpp/bindings/constraints.cpp
```diff
@@ -0,0 +1,109 @@
...
+#include <toppra/constraint/linear_joint_acceleration.hpp>
+#include <toppra/constraint/linear_joint_velocity.hpp>
+
+namespace py = pybind11;
+
+#ifdef BUILD_WITH_PINOCCHIO
+// Mind that pinocchio headers **must** be included before boost headers.
+#include <toppra/constraint/joint_torque/pinocchio.hpp>
+#ifdef PINOCCHIO_WITH_PYTHON_INTERFACE
+#include <boost/python.hpp>
```
hungpham2511, line 21: I know it works for you, but I wonder if using both boost and pybind is a good idea?

cpp/bindings/CMakeLists.txt
hungpham2511: Nevermind this comment. Since boost Python is header-only, nothing bad can happen.

hungpham2511, line 27: I don't see how PINOCCHIO_WITH_PYTHON_INTERFACE is defined?

Also, can we find different boost component based on the PYTHON_VERSION option?

jmirabel: I understand your concerns. If you reach this point in CMakeLists.txt, it means that pinocchio was installed, with the bindings so boost python is available, with a version compatible with pinocchio. If you don't have boost, you will be restricted to the `jointTorque.Pinocchio(urdfFilename)` API.
So this gives the full bindings to people who have Pinocchio Python bindings, and a restricted API to others.

jmirabel: It is defined in the `pinocchioConfig.cmake`, included when doing `find_package(pinocchio)`.

jmirabel: For Python 3, I think you can use component `python3-py36` for Python 3.6 but I am not sure.

cpp/bindings/constraints.cpp
jmirabel: I would say there is not issue, but I am not sure.

cpp/bindings/CMakeLists.txt
jmirabel: > Nevermind this comment. Since boost Python is header-only, nothing bad can happen.

It isn't header only. There are libraries.

cpp/bindings/constraints.cpp
```diff
@@ -0,0 +1,109 @@
...
+#ifdef BUILD_WITH_PINOCCHIO
+    py::module::import("pinocchio");
+    py::class_<constraint::jointTorque::Pinocchio<>,
+      std::shared_ptr<constraint::jointTorque::Pinocchio<> >,
+      constraint::JointTorque>(
+        mod_jointTorque, "Pinocchio")
+#ifdef PINOCCHIO_WITH_PYTHON_INTERFACE
+        .def(py::init([](py::object model, const toppra::Vector& fc) {
+              return toppra::constraint::jointTorque::Pinocchio<>(
+                  boost::python::extract<const pinocchio::Model&>(model.ptr()), fc);
```
jmirabel, line 100: The only potential bug is here. The Pinocchio Model is held by Python. If Python deletes the object, the joint torque constraint will point to a deleted model.

cpp/src/toppra/constraint/joint_torque/pinocchio.hpp
```diff
@@ -22,6 +23,15 @@ class Pinocchio : public JointTorque {
     typedef _Model Model;
     typedef typename _Model::Data Data;
 
+    /// Build a JointTorque constraint with no friction
+    /// \param urdfFilename path to a URDF file.
+    static Pinocchio fromURDF (const std::string& urdfFilename)
+    {
+      Model model;
+      pinocchio::urdf::buildModel(urdfFilename, model);
+      return Pinocchio(model, Vector::Zero(model.nv));
```
jmirabel, line 32: I realize now this is not correct as the return object will keep a reference to a deleted model. Please don't merge.

cpp/bindings/CMakeLists.txt
hungpham2511: I tried to find pinocchio for python3 on robotpkg, there is only the 3.6 binding. Do you say it's easy/possible to build it from source for a different version (i.e. 3.7)?

cpp/bindings/constraints.cpp
hungpham2511: model will be a copy and is owned by pybind, no? In any case it will go out-of-scope after the constructor terminates and the pointer will be invalid all the same.

hungpham2511:
I am ok with having boost.Python as an optional dependency. In fact, since pinocchio seems to be very nice, I might start to use it instead of openrave for kinematics/dynamics related computation as well.

cpp/bindings/CMakeLists.txt
jmirabel: Should be relatively easy:
```bash
sudo apt remove robotpkg-pinocchio
# (optional, for collision detection) sudo apt install robotpkg-hpp-fcl 
git clone --recursive https://github.com/stack-of-tasks/pinocchio
mkdir build && cd build
cmake -DPYTHON_EXECUTABLE=<path to python 3.7 interpreter> ..
# ccmake .. to tune the few compilation options
make install # go take a coffee, it's long and memory consuming.
```

cpp/bindings/constraints.cpp
jmirabel: Sorry for the confusion:
by model, I meant the `pInocchio::Model` object, not the `py::object`. The `py::object`, when created, increases the reference to the already existing Python object wrapping the `pinocchio::Model`. This `py::object` is temporary, so the owning is temporary too. After this method returns, the only person who owns the object is Python.

This means that
```python
def function():
  model = createModel()
  return toppra.cpp.jointTorque.Pinocchio(model, np.zeros(model.nv))

c = function()
```
will segv (or have undefined behavior) when using `c`.

jmirabel: I think pybind11 offers a way of binding the lifetime of a Python object to another object. This could fix the issue.

https://pybind11.readthedocs.io/en/master/advanced/functions.html#keep-alive

jmirabel: This is now fixed.

cpp/bindings/CMakeLists.txt
jmirabel: You can also install Pinocchio from anaconda or pip.

jmirabel:
From my side, this is ready.

cpp/src/toppra/algorithm/toppra.cpp
```diff
@@ -33,7 +33,7 @@ ReturnCode TOPPRA::computeForwardPass(value_type vel_start) {
     }
     /// \todo This can be optimized further by solving a 1D problem instead of 2D
     m_data.parametrization(i + 1) =
-        m_data.parametrization(i) + 2 * deltas(i) * solution(0);
+      std::max(0., m_data.parametrization(i) + 2 * deltas(i) * solution(0));
```
jmirabel, line 36: I highlight this change just to be sure you saw it.

hungpham2511:
Great! Many thanks.

cpp/bindings/constraints.cpp
```diff
@@ -0,0 +1,112 @@
...
+  py::class_<JointTorque,
+    std::shared_ptr<JointTorque>,
+    LinearConstraint>(
+      m, "JointTorque");
+
+  {
+    auto mod_jointTorque = m.def_submodule("jointTorque");
+#ifdef BUILD_WITH_PINOCCHIO
+    using jointTorque::Pinocchio;
+    py::module::import("pinocchio");
```
hungpham2511, line 96: @jmirabel  This line requires pinocchio to be available in the python site-packages when the user only wants to use the Pinocchio C++ library. I try to comment out and it doesn't seem to cause any issue.  Do you really need it in this case? If no I will make the below change.

`
#ifdef PINOCCHIO_WITH_PYTHON_INTERFACE
    py::module::import("pinocchio");
#endif
`

jmirabel: You are right. Your change is correct.

PR #118, opened 2020-05-31
[cpp][cmake] Fix PYTHON_VERSION variable.

jmirabel:
CMake `option` only supports boolean variable (at least for CMake 3.10).


hungpham2511:
Worked for me. Probably because I use 3.17

jmirabel:
It may have worked, but the doc says it should be a boolean value: https://cmake.org/cmake/help/v3.17/command/option.html

PR #125, opened 2020-06-11
[cpp] Add the possible to use varying joint velocity limits.

### LINKED ISSUE #124: Varying joint velocity constraints for cpp
**Is your feature request related to a problem? Please describe.**
Currently the C++ does not seem to support varying joint velocity constraints for the path (velocity per way-point) in the same manner as the python API.

It would be god to see an example of varying velocity contraints with the C++ implementation or the feature added if the API cannot support it. 

**Describe alternatives you've considered**
Currently my organization uses the library in a python context but our stack is in C++. We would like to migrate this library to C++, however the lack of varying velocity constraints in the C++ implementation is a blocker.

jmirabel:
Fixes #124 

Checklists:

### DIFF
--- a/cpp/src/toppra/constraint/linear_joint_velocity.hpp
+++ b/cpp/src/toppra/constraint/linear_joint_velocity.hpp
@@ -6,7 +6,24 @@
 namespace toppra {
 namespace constraint {
 
-/// A Joint Velocity Constraint class.
+/**
+\brief A Joint Velocity Constraint class.
+
+This class implements the case of constant velocity limits but can be derived to
+achieve varying velocity limits. E.g.
+\code
+class LinearJointVelocityVarying : public LinearJointVelocity {
+public:
+  LinearJointVelocityVarying(...) : LinearJointVelocity (nDof) {}
+protected:
+  void computeVelocityLimits(value_type time)
+  {
+    m_lower = ...;
+    m_upper = ...;
+  }
+};
+\endcode
+*/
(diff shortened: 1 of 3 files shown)


cpp/src/toppra/constraint/linear_joint_velocity.hpp
```diff
@@ -6,7 +6,24 @@
...
+public:
+  LinearJointVelocityVarying(...) : LinearJointVelocity (nDof) {}
+protected:
+  void computeVelocityLimits(value_type time)
+  {
+    m_lower = ...;
+    m_upper = ...;
+  }
+};
+\endcode
```
jmirabel, line 25: @thummel42 here is how to use this.

hungpham2511:
Fantastic. Do you think it's possible to have a default implementation that linearly interpolates some given limits?

jmirabel:
> Fantastic. Do you think it's possible to have a default implementation that linearly interpolates some given limits?

I implemented this to give the design which avoids duplicated code. Maybe @thummel42 can do what you suggest.

cpp/src/toppra/constraint/linear_joint_velocity.hpp
thummel42: Awesome thanks!! Looking forward to the merge.

hungpham2511:
Ok I will merge this PR as it is. @thummel42 If you would like to contribute your implementation of a Varying velocity constraint, please feel free to open a PR.

PR #129, opened 2020-06-24
[cpp]Implement trajectory parametrizer

### LINKED ISSUE #102: Generate output trajectory
A feature that allows to "plug" in the new time parametrization to the original geometric path and outputs:
- joint pos/vel/accel at given time instances;

This can perhaps be done by adding a another method to the GeometricPath class, e.g. 'applyParametrization', to obtain the new trajectory.

hungpham2511:
Fixes #102 

Changes in this PRs:
- [x] Constant Accel parametrizer
- [x] Unit tests

Checklists:


cpp/tests/test_toppra_approach.cpp
```diff
@@ -0,0 +1,152 @@
...
+  velLimitUpper << 1, 0.5;
+  accLimitLower << -0.05, -0.1;
+  accLimitUpper << 0.2, 0.3;
+
+  toppra::LinearConstraintPtr ljv, lja;
+  ljv = std::make_shared<toppra::constraint::LinearJointVelocity>(
+      velLimitLower, velLimitUpper);  //[[-1, 1], [-0.5, 0.5]]
+  lja = std::make_shared<toppra::constraint::LinearJointAcceleration>(
+      accLimitLower, accLimitUpper);  //[[-0.05, 0.2], [-0.1, 0.3]]
+  lja->discretizationType(toppra::DiscretizationType::Interpolation);
```
hungpham2511, line 70: @ndehio This is necessary.

hungpham2511:
@jmirabel Please let me know your comments if you have any. Thanks. 

@ndehio The approach example is working now. Aside from a few bugs in my parametrizer impl, it fails because of a numerical error. In this case, you should use Interpolation method. See another of my comments. 

In fact, I think Interpolation should be the default option and will probably make that change shortly.

cpp/src/toppra/parametrizer/const_accel.cpp
```diff
@@ -0,0 +1,129 @@
...
+    // increment k_grid until the condition is satisfied
+    while (k_grid < (m_ts.size() - 1)) {
+      if (m_ts[k_grid] <= ts[i] && ts[i] < m_ts[k_grid + 1]) {
+        break;
+      } else if (k_grid == (m_ts.size() - 2)) {
+        break;
+      } else {
+        k_grid++;
+      }
+    }
```
jmirabel, line 58: FYI, this loop can be made logarithmic with the help of either [std::lower_bound](https://en.cppreference.com/w/cpp/algorithm/lower_bound) or [std::upper_bound](https://en.cppreference.com/w/cpp/algorithm/upper_bound).

If the number of grid point is high, this can make a substencial difference.

cpp/src/toppra/parametrizer/const_accel.cpp
```diff
@@ -0,0 +1,129 @@
...
+  }
+  return true;
+}
+
+Vectors ConstAccel::eval_impl(const Vector& times, int order) const {
+  Vector ss, vs, us;
+  TOPPRA_LOG_DEBUG("eval_impl. order=" << order);
+  bool ret = evalParams(times, ss, vs, us);
+  assert(ret);
+  if (order == 0) {
```
jmirabel, line 75: I find `switch` more readable in this context.

cpp/src/toppra/toppra.hpp
```diff
@@ -17,6 +17,7 @@
 #define TOPPRA_LOG_DEBUG(X) ((void)0)
 #endif
 
+#define TOPPRA_NEARLY_ZERO 1e-8
```
jmirabel, line 20: Is this used somewhere ?

jmirabel (approved):
I made a few minor comments. This PR looks fine to me. I haven't looked at the unit-tests.

cpp/src/toppra/parametrizer/const_accel.cpp
hungpham2511: Thanks for the suggestion. It makes perfect sense. Though I am not sure how to obtain begin/end iterators from a Eigen::Matrix. Do you have a suggestion?

jmirabel: It is a vector so you should be able to do:
`std::lower_bound(m_ts.data(), m_ts.data()+m_ts.size(), value)`.

hungpham2511: https://stackoverflow.com/questions/27030550/foreach-loops-over-eigen-matrices

This suggests using x.data() and x.data() + x.size(). Tbh that looks a bit "unsafe" .

jmirabel: I think it is explicitly written somewhere in the STL documentation that you can pass raw pointer in place of iterators. This is safe.

cpp/src/toppra/toppra.hpp
hungpham2511: I use this here https://github.com/hungpham2511/toppra/pull/129/files#diff-19af0bded9ec5a7749236d4bcce703caR18

cpp/src/toppra/parametrizer/const_accel.cpp
```diff
@@ -0,0 +1,128 @@
...
+    //      k_grid = 0 if                  ts[i] < m_ts[0], or
+    //      k_grid = m_ts.size() - 2 if    ts[i] > m_ts
+    auto ptr = std::lower_bound(m_ts.data(), m_ts.data() + m_ts.size(), ts[i]);
+    if (ptr == m_ts.data() + m_ts.size()) {
+      k_grid = m_ts.size() - 2;
+    } else if (ptr == m_ts.data()) {
+      k_grid = 0;
+    } else {
+      k_grid = ptr - m_ts.data() - 1;
+    }
```
jmirabel, line 52: Optional: Isn't this equivalent ?
```suggestion
    assert (m_ts.size() >= 2);
    auto ptr = std::lower_bound(m_ts.data() + 1, m_ts.data() + m_ts.size() - 1, ts[i]);
    k_grid = ptr - m_ts.data() - 1;
```

hungpham2511: Indeed. Thanks!

cpp/tests/test_toppra_approach.cpp
ndehio: Thanks a lot for completing my pull request. What about adding also `ljv->discretizationType(toppra::DiscretizationType::Interpolation);`  ?

PR #140, opened 2020-09-13
Adding full cpp-example using parametrizer

ndehio:
Fixes #
- add missing validate() method
- assign return-code in m_data.ret_code
- add missing `parametrizer` in CMake
- add full toppra-example (note: in its current state the test fails...)

Note
I did not find a full example showing how to use the cpp-interface. The new test-file `test_toppra_approach.cpp` implements the python-example described [here](https://hungpham2511.github.io/toppra/quickstart.html). Unfortunately, it does not compute a similar optimized trajectory and fails. I don't understand the reason for this. Feedback is welcome.

### DIFF
--- a/cpp/src/CMakeLists.txt
+++ b/cpp/src/CMakeLists.txt
@@ -67,6 +67,7 @@ install(FILES
   ${CMAKE_CURRENT_BINARY_DIR}/toppra/export.hpp
 
   toppra/geometric_path.hpp
+  toppra/parametrizer.hpp
   toppra/toppra.hpp
 
   toppra/solver.hpp
@@ -83,7 +84,9 @@ install(FILES
 install(FILES
   toppra/geometric_path/piecewise_poly_path.hpp
   DESTINATION include/toppra/geometric_path)
-
+install(FILES
+  toppra/parametrizer/const_accel.hpp
+  DESTINATION include/toppra/parametrizer)
 install(FILES
   toppra/constraint/joint_torque/pinocchio.hpp
   DESTINATION include/toppra/constraint/joint_torque)
--- a/cpp/tests/test_toppra_approach.cpp
+++ b/cpp/tests/test_toppra_approach.cpp
@@ -1,0 +1,164 @@
+//Author: Niels Dehio, 13. September 2020
+#include <toppra/geometric_path/piecewise_poly_path.hpp>
+#include <toppra/constraint/linear_joint_velocity.hpp>
+#include <toppra/constraint/linear_joint_acceleration.hpp>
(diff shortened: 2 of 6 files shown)


hungpham2511:
Thanks for the PR! I will have a look.

cpp/src/CMakeLists.txt
```diff
@@ -83,7 +84,9 @@ install(FILES
 install(FILES
   toppra/geometric_path/piecewise_poly_path.hpp
   DESTINATION include/toppra/geometric_path)
-
+install(FILES
+  toppra/parametrizer/const_accel.hpp
```
jmirabel, line 88: Didn't you forget to commit this file ?

ndehio: Please remember that this PR is related to the branch `feat-parametrizer`.
The `const_accel` files were already [there.](https://github.com/hungpham2511/toppra/tree/feat-parametrizer/cpp/src/toppra/parametrizer)

jmirabel: Ok, I missed that. Thanks.

ndehio: Looking forward to find this `parametrizer` extension also in the master ;-)

cpp/tests/test_toppra_approach.cpp
```diff
@@ -0,0 +1,164 @@
...
+  if(printInfo) std::cout << "pd.ret_code = " << int(pd.ret_code) << std::endl;
+  ASSERT_EQ(pd.ret_code, toppra::ReturnCode::OK);
+
+  //#### create constant accelaration parametrizer ####
+  Eigen::VectorXd gridpoints = pd.gridpoints; //Grid-points used for solving the discretized problem.
+  Eigen::VectorXd vsquared = pd.parametrization; //Output parametrization (squared path velocity)
+  std::shared_ptr<toppra::parametrizer::ConstAccel> ca;
+  ca = std::make_shared<toppra::parametrizer::ConstAccel>(path, gridpoints, vsquared);
+  if(printInfo) std::cout << "ca->validate() = " << ca->validate() << std::endl;
+  ASSERT_FALSE(ca->validate()); //TODO why false? => ASSERT_TRUE(ca->validate());
```
hungpham2511, line 124: This should not be false. Look like there is a bug in the parametrizer class, or this instance is not feasible.

ndehio: I fully agree, but I don't understand how to fix it.

hungpham2511:
Thanks for the PR. I will look into the failure.

PR #143, opened 2020-10-07
[Cpp] Add PathParametrizationAlgorithm::setGridpoints

jmirabel:


jmirabel:
The Python 2.7 error seems unrelated to this PR.

cpp/src/toppra/algorithm.cpp
```diff
@@ -105,12 +105,28 @@ ReturnCode PathParametrizationAlgorithm::computeFeasibleSets() {
   return ret;
 }
 
+void PathParametrizationAlgorithm::setGridpoints(const Vector& gridpoints)
+{
+  Bound I (m_path->pathInterval());
+  auto s = gridpoints.size();
+  if (s != m_N + 1)
+    /// \todo shouldn't we set m_N instead ?
```
hungpham2511, line 113: Yes indeed there is no reason to fix m_N if we are setting the gridpoints directly.

cpp/src/toppra/algorithm.cpp
```diff
@@ -105,12 +105,28 @@ ReturnCode PathParametrizationAlgorithm::computeFeasibleSets() {
...
+  m_data.gridpoints = gridpoints;
+}
+
 void PathParametrizationAlgorithm::initialize() {
   if (m_initialized) return;
   if (!m_solver)
     throw std::logic_error("You must set a solver first.");
   Bound I (m_path->pathInterval());
-  m_data.gridpoints = Vector::LinSpaced(m_N + 1, I(0), I(1));
+  if (m_data.gridpoints.size() != m_N+1)
```
hungpham2511, line 128: I think we should only initialize `m_data.gridpoints` if it is not initialized before, and not when the number of gridpoints mismatch.

It is probably better now do m_data.parametrization.resize(m_data.gridpoints.size()) to make the intent that the number of gridpoints and the number of elements in the parametrization vector are the same, and thus remove the m_N usage.

cpp/src/toppra/algorithm.cpp
```diff
@@ -105,12 +105,28 @@ ReturnCode PathParametrizationAlgorithm::computeFeasibleSets() {
...
+  auto s = gridpoints.size();
+  if (s != m_N + 1)
+    /// \todo shouldn't we set m_N instead ?
+    throw std::invalid_argument("gridpoints is of wrong size.");
+  if (   std::fabs(gridpoints[0] - I(0)) > TOPPRA_NEARLY_ZERO
+      || std::fabs(gridpoints[gridpoints.size()-1] - I(1)) > TOPPRA_NEARLY_ZERO)
+    throw std::invalid_argument("Path interval does not match first and last gridpoints.");
+  if (((gridpoints.tail(s-1) - gridpoints.head(s-1)).array() <= 0).any())
+    throw std::invalid_argument("gridpoints should be strictly increasing.");
+  m_data.gridpoints = gridpoints;
```
hungpham2511, line 120:
```suggestion
  m_data.gridpoints = gridpoints;
  m_initialized = False;
```

hungpham2511 (approved):
LGTM! Can you also update the HISTORY.md file?

jmirabel:
Thanks.

Issue #101, opened 2020-05-07
Cartesian acceleration/velocity constraints 


hungpham2511:
**Describe the solution you'd like**
Implementations of Cartesian acceleration/velocity constraints with the C++ API

Now it's possible to implement this using either OpenRAVE or https://github.com/stack-of-tasks/pinocchio Probably we should implement both.

An example interface:
```C++

```

thummel42:
Hey just asking if there is an example python interface that can be ported to cpp for this. Specifically would like to do varying velocity and acceleration constraints.

hungpham2511:
Yes I plan to do this for sometime for unfortunately couldn't find the time for it. It should be a straightforward extension of https://github.com/hungpham2511/toppra/blob/develop/toppra/constraint/linear_second_order.py

stevegolton:
Hi @hungpham2511. I had a go at making a Cartesian velocity constraint class (see listing below). It seems to work just fine using an `fk` implementation based on PyKDL's `ChainFkSolverVel_recursive`, but I'm not sure why I have to square the outputs of the `invdyn` and `constraintg` functions. I'll admit my understanding of the maths is hazy. Is this something like the class you had imagined?

```
class CartesianVelocityConstraint(constraint.SecondOrderConstraint):
    """This class implements a constraint on Cartesian linear and angular velocity of some link in the robot chain.
    """

    def __init__(self, fk, max_linear, max_angular, dof):
        """Initialize the constraint.

        Parameters
        ----------
        fk: (np.ndarray, np.ndarray) -> (linear_vel, angular_vel)
            The "FK" function that receives joint position and velocity as inputs and ouputs
            the linear and angular velocity of some link of the robot.
        max_linear: float
            The max linear velocity allowed for that link.
        max_angular: float
            The max angular velocity allowed for that link.
        dof: int
            The dimension of the joint position.
        """
        super(CartesianVelocityConstraint, self).__init__(self.invdyn, self.constraintf, self.constraintg, dof)
        self.fk = fk
        self.max_linear = max_linear
        self.max_angular = max_angular

    def invdyn(self, q, dq, ddq):
        linear, angular = self.fk(q, dq)
        return np.array([linear**2, angular**2])

    def constraintf(self, q):
        # 2x2 identity matrix
        return np.identity(2)

    def constraintg(self, q):
        return np.array([self.max_linear**2, self.max_angular**2])
```

hungpham2511:
[...]

PR #144, opened 2020-10-12
Added CartesianSpeedConstraint

stevegolton:
Added CartesianSpeedConstraint to constrain the magnitude of the Cartesian velocity vectors for a part of the robot.

Changes in this PR:
- Added a new class CartesianSpeedConstraint in constraint/cartesian_constraints.py
- Added example in examples/plot_cartesian_constraints.py

Checklists:
- [ ] Update CHANGELOG.md with a single line describing this PR - I was unable to find this file?

### DIFF
--- a/toppra/constraint/cartesian_constraints.py
+++ b/toppra/constraint/cartesian_constraints.py
@@ -2,7 +2,13 @@ import numpy as np
 from .linear_second_order import SecondOrderConstraint
 
 class CartesianSpeedConstraint(SecondOrderConstraint):
-    """This class implements a constraint on the linear & angular Cartesian speed of some part of the robot.
+    """
+    This class implements a constraint on the magnitudes of the linear & angular
+    Cartesian velocity vectors of one of the robot's parts (link, joint, etc).
+    
+    The forward kinematic velocity is be provided via a callback, which makes
+    this constraint agnostic of the robot's geometry and agnostic of whatever
+    FK algorithm is used.
     """
 
     def __init__(self, fk, linear_speed_max, angular_speed_max, dof):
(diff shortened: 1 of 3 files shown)


toppra/constraint/cartesian_constraints.py
```diff
@@ -0,0 +1,43 @@
...
+            The dimension of the joint position.
+        """
+        super(CartesianSpeedConstraint, self).__init__(self.invdyn, self.constraintf, self.constraintg, dof)
+        self.fk = fk
+        self.linear_speed_max = linear_speed_max
+        self.angular_speed_max = angular_speed_max
+
+    def invdyn(self, q, dq, ddq):
+        linear_speed, angular_speed = self.fk(q, dq)
+        return np.array([linear_speed**2, angular_speed**2])
```
hungpham2511, line 37: This seems strange. We shouldn't have to square the speeds nor the speed limits.

hungpham2511:
Hi @stevegolton, many thanks for the PR. I am sorry for not getting back earlier. Was really busy at work in the past weeks. :) 

Could you help to make the CI task call `cpp-python36`? Look like PyKDL was used but not installed.

> Update CHANGELOG.md with a single line describing this PR - I was unable to find this file?

It should be HISTORY.md

toppra/constraint/cartesian_constraints.py
stevegolton: I thought it was weird too. If I don't square then it doesn't seem to work at all. The below screenshots should both have a speed limits of 1m/s and 1rad/s.

Without squaring

With squaring

stevegolton:
No worries! Likewise, apologies for not getting back to you sooner either. Work was really busy for me too.

I'm struggling a little with the PyKDL dependency issue. I was using this library because I am used to its C++ API, but in hindsight it looks like it isn't well supported and I'm having a bunch of issues getting the python library installed on the `cimg/python:3.6` docker container.

I think I might update the example to use pinocchio instead.

toppra/constraint/cartesian_constraints.py
jmirabel: I think you have to square it for the reason I mentioned in #101. Following the derivation of #146, you should be able to understand why.

hungpham2511:
I am sure that you don't need to square the cartesian acceleration. Most likely there is a bug somewhere.

Bests,

Hung

________________________________

hungpham2511:
Ah, the constraints are on Cartesian velocity! That's why. The way that you implement it only work if the constraints are second order but velocity constraints are first-order.

Bests,

Hung

________________________________

toppra/constraint/cartesian_constraints.py
stevegolton: Thanks @jmirabel, I think I've got my head around it now

stevegolton:
Looks like #146  does everything this PR does and more (c++ interface & variable velocity limits) which is great! Thus, there's no need for this PR anymore so I'll close it.

mrunaljsarvaiya:
@stevegolton Did you land up using the cpp-binding to use the cartesian constraints in python or write a pure python constraint using the cpp code as a reference?

stevegolton:
@mrunaljsarvaiya Hey, I ended up using the cpp version natively in the end. I haven't touched the python versions or any of the bindings since this PR.

PR #146, opened 2020-10-28
[cpp] Add CartesianVelocityNorm and an implemention with Pinocchio

### LINKED ISSUE #101: Cartesian acceleration/velocity constraints

jmirabel:
Partially fixes #101 and is related to #144

From the doxygen doc:
image


jmirabel:
@hungpham2511 let me know if I should care of the two CI failures.

hungpham2511:
Great! Flawless work as always. Many thanks! @jmirabel 

I wonder if S should be constrained to being positive-definite.

hungpham2511:
Ah, my bad for not reading the description of S and J. It will always be semi-positive definite at least then.

> @hungpham2511 let me know if I should care of the two CI failures.

These two are deprecated. I will remote them shortly.

hungpham2511 (approved):

hungpham2511:
By the way, Joseph, would you like to be the [code owner](https://docs.github.com/en/free-pro-team@latest/github/creating-cloning-and-archiving-repositories/about-code-owners) for the cpp codebase? Basically, if there are PRs touching the cpp code, you will get notified and invited to review.

jmirabel:
> By the way, Joseph, would you like to be the [code owner](https://docs.github.com/en/free-pro-team@latest/github/creating-cloning-and-archiving-repositories/about-code-owners) for the cpp codebase? Basically, if there are PRs touching the cpp code, you will get notified and invited to review.

At the moment, I prefer to leave it as it is. I don't always get the time necessary for a correct review and prefer not to block anyone.

mrunaljsarvaiya:
Does anyone happen to have a python example for this?

jmirabel:
Check the C++ unit tests. They should be straightforward to convert into python code.

You can open a PR with an example if you have something that works.

mrunaljsarvaiya:
> Check the C++ unit tests. They should be straightforward to convert into python code.
> 
> You can open a PR with an example if you have something that works.

I'm a newbie to python bindings, how would I import the `CartesianVelocityNorm` constraint?

After `import toppra.cpp as tac`, `tac.bindings_loaded()` always returns False.
 
`from toppra.cpp.toppra_int import *` returns an error `ModuleNotFoundError: No module named 'toppra.cpp.toppra_int'
`

jmirabel:
You haven't installed the C++ bindings then. Install them and try again.

mrunaljsarvaiya:
Yeah that's correct, I managed to install the python bindings and am able to import `CartesianVelocityNorm`. Based on the cpp test as an example, is the next step to build a class that derives from `CartesianVelocityNorm` and define the `computeVelocity` and `computeVelocityLimit` method? 

Also since the init header is `CartesianVelocityNorm (const Matrix& S, const double& limit)` in the cpp file, in python would I initialize it with the inputs being an numpy ndarray and float?

Or after taking a closer look at the bindings/constraint.cpp file, the python interface  method would be `py::init([](py::object model, const Matrix& S, const double& limit, int frame_id)`, i.e I need a pinocchio model of my system?

mrunaljsarvaiya:
If `BUILD_WITH_PINOCCHIO=True` and `PINOCCHIO_WITH_PYTHON_INTERFACE=False`, the `constraints.cpp` file will build the following py functions for `jointTorque`
`py::class_<Pinocchio<>, std::shared_ptr<Pinocchio<> >, JointTorque>(
        mod_jointTorque, "Pinocchio")
        .def(py::init<const std::string&, const Vector&>(),
            py::arg("urdfFilename"), py::arg("frictionCoeffs") = Vector())`

however, the equivalent condition for `cartesianVelocityNorm` builds the following

`py::class_<Pinocchio<>, std::shared_ptr<Pinocchio<> >, CartesianVelocityNorm>(
        mod_jointTorque, "Pinocchio")`

Is there a `.def(py.. .` missing? 

Apologies for the stream of questions here

jmirabel:
You can't create objects of type `cartesianVelovityNorm::Pinocchio` in Python without Pinocchio python bindings.

I suggest you install Pinocchio Python bindings and recompile with the relevant CMake flags. This is the easiest way.

rajkumar2804:
Do we have any  constant Cartesian speed (magnitude of velocity) implementation example , I looked into the test case but that doesn't help much.

jmirabel:
@mrunaljsarvaiya may have an example.

The first thing you should do is to make sure you have pinocchio installed with python bindings.
Then compile this project with the relevant cmake flags (you can use ccmake or another cmake gui of your choice).

PR #160, opened 2021-01-22
[cpp] Enhance Seidel solver.

jmirabel:
Changes in this PRs:
- Make Seidel solver more stable numerically.
- Add PathParameterizatonAlgorithm::setInitialXBounds

Checklists:


cpp/src/toppra/solver/seidel.cpp
```diff
@@ -195,31 +77,59 @@ LpSol solve_lp2d(const RowVector2& v,
   // adhered to: fixed bounds are assigned the numbers: -1, -2, -3,
   // -4 according to the following order: low[0], high[0], low[1],
   // high[1].
-  if ((low.array() > high.array()).any()) {
-    TOPPRA_LOG_DEBUG("Seidel: incoherent bounds.");
-    return INFEASIBLE;
-  }
   for (int i = 0; i < 2; ++i) {
+    if (low[i] > high[i]) {
```
hungpham2511, line 81: I am not sure I understand the purpose of this section. If there is an i such that low[i] and high[i], shouldn't the problem be declared infeasible?

hungpham2511 (commented):
If I understand correctly, the enhancement is on the source code, am I right?

I couldn't find any improvement that relates to numerical stability (except the comment I had in lp_solve2d). As I mentioned before, if you experience numerical instability, it is perhaps better to consider scaling the 2d lp problem before solving.

jmirabel (commented):
> If I understand correctly, the enhancement is on the source code, am I right?

Yes, you are right. 

> I couldn't find any improvement that relates to numerical stability (except the comment I had in lp_solve2d).

The issues I encountered are the following.
- In `solve_lp1d`,
   - the hard-coded bounds made the problem infeasible while it clearly was,
   - and the lower bound sometimes become greater than the upper bound by a relatively very tiny amount.
- in `solve_lp2d`,
   - I had bound issues as in `solve_lp1d`,
   - and I optimized the code to avoid
     - the duplicated checks on the lp1d inputs of lp1d function,
     - apply the index map before solving the LP (avoids memory jumps that are hard to predict),
- There was a lot of parameters left for manual tuning. Now there are only two: an absolute tolerance and a relative tolerance.

> As I mentioned before, if you experience numerical instability, it is perhaps better to consider scaling the 2d lp problem before solving.

This code cannot solve all the numerical issues and scaling would help too. However, a 2D Seidel LP solver is sufficiently simple for addressing some of the numerical issues internally. I have added a unit test for each of numerical issue I encountered.

I have added some comments. Let me know what you want me to do.

hungpham2511 (approved):
I have no further comments. Happy to merge as-is.

jmirabel:
Thanks for the review.

PR #162, opened 2021-02-28
[CPP] Implement Spline Parametrizer

leonardoedgar:
Changes in this PRs:
- Spline Parametrizer 

Checklists:

### DIFF
--- a/cpp/src/toppra/parametrizer/spline.cpp
+++ b/cpp/src/toppra/parametrizer/spline.cpp
@@ -1,0 +1,61 @@
+#include <toppra/parametrizer/spline.hpp>
+#include <toppra/geometric_path/piecewise_poly_path.hpp>
+
+namespace toppra {
+
+namespace parametrizer {
+
+Spline::Spline(GeometricPathPtr path, const Vector& gridpoints, const Vector& vsquared)
+    : Parametrizer(path, gridpoints, vsquared) {
+        std::vector<value_type> t_grid (gridpoints.rows(), 0);
+        std::vector<value_type> copied_gridpoints (gridpoints.data(), gridpoints.data() + gridpoints.rows());
+        std::vector<int> skip_ent;
+        value_type sd_average, delta_s, delta_t;
+        for (int i = 1; i < t_grid.size(); i++) {
+            sd_average = (m_vs[i - 1] + m_vs[i]) / 2;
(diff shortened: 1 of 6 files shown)


cpp/src/toppra/parametrizer/spline.cpp
```diff
@@ -0,0 +1,60 @@
...
+        std::vector<value_type> copied_gridpoints (gridpoints.data(), gridpoints.data() + gridpoints.rows());
+        std::vector<int> skip_ent;
+        value_type sd_average, delta_t;
+        for (int i = 1; i < t_grid.size(); i++) {
+            sd_average = (m_vs[i - 1] + m_vs[i]) / 2;
+            if (sd_average > TOPPRA_NEARLY_ZERO) {
+                delta_t = (gridpoints[i] - gridpoints[i - 1]) / sd_average;
+            }
+            else {
+                delta_t = 5;
```
leonardoedgar, line 20: Currently, given input squared path velocity = 0, the class will handle it with this, taking reference from the Python API https://github.com/hungpham2511/toppra/blob/facf380976e736ee01f657de7a507f5932cf4fba/toppra/parametrizer.py#L181

Thus, the parametrization validate method will always return True in this case.

For negative path squared velocity input, it is already handled in the base class https://github.com/hungpham2511/toppra/blob/facf380976e736ee01f657de7a507f5932cf4fba/cpp/src/toppra/parametrizer.cpp#L12

cpp/src/toppra/parametrizer/spline.cpp
```diff
@@ -0,0 +1,60 @@
...
+        final_bc.values = m_path->eval_single(path_interval[1], 1) * m_vs[m_vs.rows() - 1];
+        m_path = std::make_shared<PiecewisePolyPath>(q_grid, m_ts, std::array<BoundaryCond, 2>{init_bc, final_bc});
+}
+
+Vectors Spline::eval_impl(const Vector &times, int order) const {
+    return m_path->eval(times, order);
+}
+
+bool Spline::validate_impl() const {
+    return true;
```
leonardoedgar, line 49: Note that currently in this class, the validate method always returns True

hungpham2511 (approved):
Look pretty good to me.

hungpham2511:
Thanks!

Issue #156, opened 2021-01-07
[CPP] Limit on the sd variable


jmirabel:
Dear @hungpham2511,

is there a rationale for the hardcoded values
```c++
  x << 0, 100;
  x_next << 0, 1;
```
at line https://github.com/hungpham2511/toppra/blob/ec723091a717373413d5747d6c1e56d57a646fa1/cpp/src/toppra/algorithm.cpp#L40 ?

I couldn't find such a limit in the Python implementation.

hungpham2511:
You are right, there is no reason particular reason for that upper limit to be there.

However, we have had problems in the past when x is too large, the optimizer behaves quite badly. In the Python implementation, there is a scaling step to address this problem. See https://github.com/hungpham2511/toppra/blob/644e263269e45be8fc595bae59e959b9570edaa3/toppra/solverwrapper/hot_qpoases_solverwrapper.py#L120

This step should, however, be implemented by the solver wrappers like in the Python implementation because solvers are generally different when it comes to numerical instability.

jmirabel:
Thank you. I added a comment in the code.
5099b23f06acb28e822e99e65dbda08c886b69d7

PR #163, opened 2021-03-08
[CPP] Invented a parallel algorithm based on seidel implementation. 

JS00000:
Fixes #
PathParametrizationAlgorithm::computeFeasibleSets() now can pass the test.

Changes in this PRs:
- add class SeidelParallel
- Modify the class PathParametrizationAlgorithm and class Solver for parallel algorithm

Checklists:
- [ ] Update HISTORY.md with a single line describing this PR


hungpham2511:
Cool. Thanks for contributing @JS00000. I will try to get around to review the code quickly.

hungpham2511:
Do you have any numbers/graphs demonstrating the new algorithm?

JS00000:
> Do you have any numbers/graphs demonstrating the new algorithm?

Not yet. The idea of this parallel optimization is very simple, that is, separate the parts that have no dependencies in the serial calculation of the controllable set, and then process them in parallel.

cpp/src/toppra/algorithm.cpp
```diff
@@ -13,15 +23,24 @@ PathParametrizationAlgorithm::PathParametrizationAlgorithm(
 
 ReturnCode PathParametrizationAlgorithm::computePathParametrization(value_type vel_start,
                                                                     value_type vel_end) {
+  size_t be = GetCurrentTimeUsec();
   initialize();
+  size_t en = GetCurrentTimeUsec();
+  printf("time of init:                             %fms\n", (en - be)/1000.0);
+  be = en;
```
jmirabel, line 30: Can you replace `printf` by `TOPPRA_LOG_DEBUG`, here and everywhere in the code ?

cpp/src/toppra/algorithm.cpp
```diff
@@ -5,6 +5,16 @@
...
 
+#include <ctime>
+#include <sys/time.h>
+
+unsigned long long GetCurrentTimeUsec()
+{
+  struct timeval tv;
+  gettimeofday(&tv,NULL);
+  return ((unsigned long long)tv.tv_sec * 1000000 + (unsigned long long)tv.tv_usec);
+}
```
jmirabel, line 16: This is a C-style way of measuring time. Although valid, I would be in favor of a more C++ style (e.g. https://www.techiedelight.com/measure-elapsed-time-program-chrono-library).
I think `std::chrono` is also more portable.

cpp/src/toppra/solver/seidel-parallel.cpp
```diff
@@ -0,0 +1,402 @@
...
+      && Vars::ColsAtCompileTime == 1
+      && Coeffs::ColsAtCompileTime == Vars::RowsAtCompileTime + 1,
+      "Size mismatch between coefficient (1x(N+1)) and vars (Nx1)");
+  typename Coeffs::Scalar res = coeffs(coeffs.cols()-1);
+  for (int i = 0; i < Vars::RowsAtCompileTime; ++i)
+    res += (coeffs[i]==0 && !std::isfinite(vars[i]) ? 0 : coeffs[i]*vars[i]);
+  return res;
+}
+
+constexpr value_type infi = 1e6;
```
jmirabel: This *infinity* is definitely too low and I don't think this is actually necessary. What issue did you get that made you set this ?

jmirabel (commented):
Thanks for raising this point. I need a bit more information to understand your changes.

1. Duplicating the API to achieve parallelization is a bit excessive. The current API does not support parallelization. I would be in favor of updating the API to have only one API that supports parallelization, rather than changing it *opportunistically*.

2. I may be overlooking, but Seidel and SeidelParallel seems to have a lot of code in common. Is this code duplication necessary ? This is bad because:
- it makes code review hard because differences are hidden,
- it makes maintenance harder.
From what I understand, you want to store the result as an attribute of the class. I guess you also removed the reordering of the LP constraints. Am I missing something ?

I also made a few minor comments.

cpp/src/toppra/algorithm.cpp
JS00000: Sure.

cpp/src/toppra/algorithm.cpp
JS00000: Ok, It's a good suggestion.

cpp/src/toppra/solver/seidel-parallel.cpp
JS00000: If I set infi to a very large number, it will cause some numerical calculation problems. It may be caused by the limitations of the algorithm itself. In fact, the original seidel algorithm also set an upper limit of 100, which is much smaller than the upper limit here.

JS00000:
> Thanks for raising this point. I need a bit more information to understand your changes.
> 
> 1. Duplicating the API to achieve parallelization is a bit excessive. The current API does not support parallelization. I would be in favor of updating the API to have only one API that supports parallelization, rather than changing it _opportunistically_.
> 2. I may be overlooking, but Seidel and SeidelParallel seems to have a lot of code in common. Is this code duplication necessary ? This is bad because:
> 
> * it makes code review hard because differences are hidden,
> * it makes maintenance harder.
>   From what I understand, you want to store the result as an attribute of the class. I guess you also removed the reordering of the LP constraints. Am I missing something ?
> 
> I also made a few minor comments.

I will try to optimize the code structure, which will take a while.

jmirabel:
The API modification can be discussed before you actually do them. I will let @hungpham2511 give his opinion on this.

cpp/src/toppra/solver/seidel-parallel.cpp
jmirabel: In the C++ implementation, I don't think there is any such limit. Are you talking about the limit discussed in #156 ?
There is no good reason for this limit and this limit is modifiable by the user since #160.
Through this API, I change the bound to [ 0, +inf ] in my code and it works well on my use cases.

hungpham2511:
> > Do you have any numbers/graphs demonstrating the new algorithm?
> 
> Not yet. The idea of this parallel optimization is very simple, that is, separate the parts that have no dependencies in the serial calculation of the controllable set, and then process them in parallel.

To be honest, I believe there is not so much gain we can get from parallelizing the algorithm. At best solving LP(s) is two times faster. Including the time taken for other parts, i.e., calculating the coefficients, the total speedup is probably less than that.

cpp/CMakeLists.txt
```diff
@@ -17,7 +17,7 @@ option(TOPPRA_DEBUG_ON "Set logging mode to on." OFF)
 option(TOPPRA_WARN_ON "Enable warning messages." ON)
 option(OPT_MSGPACK "Serialization using msgpack " OFF)
 # Bindings
-option(PYTHON_BINDINGS "Build bindings for Python" ON)
+option(PYTHON_BINDINGS "Build bindings for Python" OFF)
```
hungpham2511: Please don't change the default flag needlessly. You can always use `-DPython_BINDINGS=OFF` during compilation.

cpp/CMakeLists.txt
```diff
@@ -47,6 +47,13 @@ if(BUILD_WITH_GLPK)
     message(STATUS "Found glpk ${GLPK_LIBRARY} ${GLPK_INCLUDE_DIR}")
 endif(BUILD_WITH_GLPK)
 
+find_package(OpenMP REQUIRED)
```
hungpham2511, line 50: It's better to have a BUILD_WITH_OPENMP flag and find OpenMP only if it is switched on. Otherwise this breaks compilation.

cpp/src/toppra/algorithm.cpp
```diff
@@ -67,6 +86,100 @@ ReturnCode PathParametrizationAlgorithm::computeControllableSets(
...
+  m_data.controllable_sets.row(m_N) = vel_ends.cwiseAbs2();
+
+  size_t be = GetCurrentTimeUsec();
+
+  bool batch_ok = true;
+  printf("omp omp_get_max_threads %d\n", omp_get_max_threads());
+  #pragma omp parallel for schedule(dynamic)
+  for (int i = 0; i < m_N; i++) {
+    // solver_ret = m_solver->solveStagewiseBatch(i, g_upper, solution_upper[i]);
+    solver_ret = m_solver->solveStagewiseBatch(i, g_upper);
```
hungpham2511, line 136: This doesn't look right to me. Are you sure that this routine returns the correct controllable sets?

hungpham2511 (commented):
After going through the change, I am actually confused about where parallelism is implemented and how does it help. Could you @JS0000 explain briefly? In particular, it is the seidel solver that you are improving or the high-level steps in the TOPPRA algorithm.

I think only by understanding how parallelism is used, we can come up with a suitable set of APIs.

cpp/src/toppra/solver/seidel-parallel.cpp
JS00000: You are right. In the origin develop branch, if you set m_initXBound to [ 0, +inf ], it will works well. But it doesn't work in parallel branch here. This is also the limitation of this parallel algorithm.
There are two variables, u and x. In principle, the limit of x (which is m_initXBound in origin develop branch) should be set to [ 0, +inf ], and the limit of u should be set to [ -inf, +inf ]. 
However, after a series of incomplete tests, I found that in parallel branch here, the limit of x can be set to [ 0, +inf ], but the limit of u should be set to [ -1e13, +1e13 ] or smaller. Otherwise, this parallel algorithm will have a large numerical error. 
The reason is that the order of constraints in the LP in this parallel algorithm has been adjusted. In order to process unrelated constraints in parallel, I put the two constraints corresponding to xNext at the end. This leads to a high degree of linear correlation between the constraints during the first and last stages of parallel processing, and the difference between the variables u and x is even up to 15 orders of magnitude. This is enough to make the "double“ data structure produce enough errors to affect the result.
So I set infi to 1e6, it can also be set larger, of course, a more complete test is needed. A smaller infi setting may speed up the calculation speed, and it is also sufficient to meet actual data requirements.

cpp/CMakeLists.txt
JS00000: OK.

cpp/CMakeLists.txt
JS00000: OK.

cpp/src/toppra/algorithm.cpp
JS00000: No, it just store the half-finished solution in solver.

JS00000:
> > > Do you have any numbers/graphs demonstrating the new algorithm?
> > 
> > 
> > Not yet. The idea of this parallel optimization is very simple, that is, separate the parts that have no dependencies in the serial calculation of the controllable set, and then process them in parallel.
> 
> To be honest, I believe there is not so much gain we can get from parallelizing the algorithm. At best solving LP(s) is two times faster. Including the time taken for other parts, i.e., calculating the coefficients, the total speedup is probably less than that.

Yes. In my tests, the total speed is atmost 2 times faster than origin Seidel solver, but still slower than qpOASES solver.

cpp/CMakeLists.txt
JS00000: And here is actually the implementation of parallelism.

cpp/src/toppra/algorithm.cpp
JS00000: And here is actually the implementation of parallelism.

Issue #151, opened 2020-12-17
Build error installing from pypi


tykurtz:
**Describe the bug**
For projects that include `toppra` as a dependency, it is necessary to pre-install `numpy` and `cython`
. 
e.g. `pip install numpy cython; pip install -r requirements.txt`

Some reading material on this issue since it's not unique to `toppra`
* https://pip.pypa.io/en/stable/reference/pip/#build-system-interface
* https://www.python.org/dev/peps/pep-0518/
* https://stackoverflow.com/a/54138355

**To Reproduce**
```sh
rm -rf ~/.cache/pip
mkvirtualenv --python 3.7 toppra-testing
pip install --upgrade pip
pip install toppra
```
which returns
```sh
    ERROR: Command errored out with exit status 1:
     command: /home/tyler/.virtual_envs/toppra-testing/bin/python -c 'import sys, setuptools, tokenize; sys.argv[0] = '"'"'/tmp/pip-install-2vcf69u8/toppra_6dbedb1d31e145f3bddb098516f7c839/setup.py'"'"'; __file__='"'"'/tmp/pip-install-2vcf69u8/toppra_6dbedb1d31e145f3bddb098516f7c839/setup.py'"'"';f=getattr(tokenize, '"'"'open'"'"', open)(__file__);code=f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' egg_info --egg-base /tmp/pip-pip-egg-info-_16s67c0
         cwd: /tmp/pip-install-2vcf69u8/toppra_6dbedb1d31e145f3bddb098516f7c839/
    Complete output (5 lines):
    Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "/tmp/pip-install-2vcf69u8/toppra_6dbedb1d31e145f3bddb098516f7c839/setup.py", line 3, in <module>
        from Cython.Distutils import build_ext
    ModuleNotFoundError: No module named 'Cython'
    ----------------------------------------
ERROR: Command errored out with exit status 1: python setup.py egg_info Check the logs for full command output.
```

**Expected behavior**
`pip install -r requirements.txt` works without pre-installing `numpy` and `cython` where requirements.txt includes `toppra`.

**Screenshots**

**Version**
Repeatable on `pypi` as well as locally on `origin/develop`

tykurtz:
Hey @hungpham2511 , thanks for merging that fix. 

[...]

PR #167, opened 2021-04-09
Use oldest supported numpy version during build

tykurtz:
Fixes an issue related to #151 . 

To reproduce

```sh
mkvirtualenv temp_venv --python=python3.7
pip install toppra==0.4.0 numpy==1.18.5  # or pip install -e . on upstream/master
python -c "import toppra"
```

and you should get

```sh
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/home/tyler/.virtual_envs/bar/lib/python3.7/site-packages/toppra/__init__.py", line 16, in <module>
    from . import constraint
  File "/home/tyler/.virtual_envs/bar/lib/python3.7/site-packages/toppra/constraint/__init__.py", line 83, in <module>
    from .linear_joint_velocity import (
  File "/home/tyler/.virtual_envs/bar/lib/python3.7/site-packages/toppra/constraint/linear_joint_velocity.py", line 3, in <module>
    from toppra._CythonUtils import (_create_velocity_constraint,
  File "toppra/_CythonUtils.pyx", line 1, in init toppra._CythonUtils
ValueError: numpy.ndarray size changed, may indicate binary incompatibility. Expected 88 from C header, got 80 from PyObject
```

Changes in this PRs:
- Fixes a build issue against older versions of numpy

Checklists:

Some background on what exactly this change is can be found here, https://pypi.org/project/oldest-supported-numpy/

> The reason to use the oldest available Numpy version as a build-time dependency is because of ABI compatibility. Binaries compiled with old Numpy versions are binary compatible with newer Numpy versions, but not vice versa.

and https://github.com/scikit-learn-contrib/hdbscan/issues/457#issuecomment-773671043

> Probably, pip has always been doing this and the only reason it's breaking now is because numpy==1.20.0 is a breaking ABI change. This also means that there's nothing to "roll back", because the mere existence of a new numpy is enough to cause it.

The other way around this would be to require `numpy>=1.20` but that would rule out setups which require older versions of `numpy` (notably `tensorflow`)

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ## Unreleased
+- [python] Fixed build error for numpy versions <= 1.20
 
 
 ## 0.4.0 (Mar 16 2021)
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,2 +1,3 @@
 [build-system]
-requires = ["setuptools", "wheel", "Cython", "numpy"]
+requires = ["setuptools", "wheel", "Cython", "oldest-supported-numpy"]
+


dgoel:
Friendly ping to @hungpham2511 for review.

hungpham2511:
Hey thanks for the PR and, pinging me. Have been kind of busy recently and sort lost track of this PR. I will have a look shortly :)

hungpham2511 (approved):
LGTM. Thanks!

dgoel:
> LGTM. Thanks!

Thank you so much! Could you publish a new release to pypi ?

hungpham2511:
I just kick-start the CI. A new version should be available soon.

PR #178, opened 2021-08-08
Allow to build static library

ahoarau:
This removes the hardcoded SHARED library build, allowing users to choose between static or shared libraries, passing -DBUILD_SHARED_LIBS=True or False.

### DIFF
--- a/cpp/src/CMakeLists.txt
+++ b/cpp/src/CMakeLists.txt
@@ -29,6 +29,8 @@ if(BUILD_WITH_GLPK)
 endif()
 
 add_library(toppra ${SOURCE_FILES})
+set_property(TARGET toppra PROPERTY POSITION_INDEPENDENT_CODE ON)
+
 FILE(MAKE_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/toppra)
 generate_export_header(toppra
   EXPORT_FILE_NAME toppra/export.hpp
--- a/cpp/src/CMakeLists.txt
+++ b/cpp/src/CMakeLists.txt
@@ -28,7 +28,7 @@ if(BUILD_WITH_GLPK)
         toppra/solver/glpk-wrapper.cpp)
 endif()
 
-add_library(toppra SHARED ${SOURCE_FILES})
+add_library(toppra ${SOURCE_FILES})
 FILE(MAKE_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/toppra)
 generate_export_header(toppra
   EXPORT_FILE_NAME toppra/export.hpp
--- a/cpp/CMakeLists.txt
+++ b/cpp/CMakeLists.txt
@@ -13,9 +13,10 @@ option(BUILD_WITH_PINOCCHIO "Compile with Pinocchio library" OFF)
 option(BUILD_WITH_qpOASES "Compile the wrapper for qpOASES" OFF)
 option(BUILD_WITH_GLPK "Compile the wrapper for GLPK (GPL license)" OFF)
 # General options
-option(TOPPRA_DEBUG_ON "Set logging mode to on." OFF)
-option(TOPPRA_WARN_ON "Enable warning messages." ON)
-option(OPT_MSGPACK "Serialization using msgpack " OFF)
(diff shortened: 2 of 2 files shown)


jmirabel:
I think this changes the default behavior which isn't desired. The doc says:

> BUILD_SHARED_LIBS
> -----------------
> 
> Global flag to cause ``add_library()`` to create shared libraries if on.
> 
> If present and true, this will cause all libraries to be built shared
> unless the library was explicitly added as a static library.  This
> variable is often added to projects as an ``option()`` so that each user
> of a project can decide if they want to build the project using shared or
> static libraries.



Can you also add
```cmake
option(BUILD_SHARED_LIBS 'Toggle between shared and static libs` TRUE)
```
so that the default is `SHARED` ?

jmirabel (changes requested):
Thanks for contributing to the windows build.

Maybe it would be good to setup CI on windows to make sure future changes won't break your work. @ahoarau any experience on this ?

hungpham2511:
Setting up CI on Windows is a great idea. If I am not wrong many of the library users are on Windows. 

FYI, I am testing out github action and it has Windows-based runners. But I am not sure how much work is needed to get it to work as I have very minimal experience compiling stuff on Windows.

jmirabel:
To be clear, the change I request is the default library type. CI on Windows is a different topic that goes into another PR.

It shouldn't be hard to implement either with github action or appveyor. @ahoarau let us know if you are wiling to do this.

ahoarau:
Basically all we need is those lines to be integrated in the CI script : 


```
git clone https://github.com/hungpham2511/toppra
cd toppra/cpp
cmake -B build-release -DPYTHON_BINDINGS=OFF
cmake --build build-release --target install --config Release -j $(nproc)
```

I wont be much help regarding the script unfortunately.

ahoarau:
Any news regarding this ?

jmirabel:
Putting aside the discussion regarding CI on Windows, you still need to make shared lib the default. Please, add
`option(BUILD_SHARED_LIBS 'Toggle between shared and static libs` TRUE)`

When building static library, it isn't sufficient to do `add_library(target STATIC ...)`. You also have to do
`set_property(TARGET target PROPERTY POSITION_INDEPENDENT_CODE)` (on linux at least). Otherwise you get the kind of error that makes CI crash:
> /usr/bin/ld: ../src/libtoppra.a(constraint.cpp.o): relocation R_X86_64_PC32 against symbol `_ZTISt16invalid_argument@@GLIBCXX_3.4' can not be used when making a shared object; recompile with -fPIC

ahoarau:
done !

cpp/src/CMakeLists.txt
```diff
@@ -28,7 +28,9 @@ if(BUILD_WITH_GLPK)
         toppra/solver/glpk-wrapper.cpp)
 endif()
 
-add_library(toppra SHARED ${SOURCE_FILES})
+add_library(toppra ${SOURCE_FILES})
+set_property(TARGET toppra PROPERTY POSITION_INDEPENDENT_CODE ON)
```
jmirabel, line 32: Does it do something when building shared libs ? By security, I would wrap it in an IF(NOT BUILD_SHARED_LIBS).

jmirabel (changes requested):
Thanks. We are already there. Sorry for insisting but you haven't addressed this :

> Can you also add
> ```cmake
> option(BUILD_SHARED_LIBS 'Toggle between shared and static libs` TRUE)
> ```
> so that the default is `SHARED` ?

cpp/src/CMakeLists.txt
ahoarau: It's fine to keep it when we build shared libs.

jmirabel: From [CMake doc](https://cmake.org/cmake/help/latest/command/add_library.html):
> For SHARED and MODULE libraries the POSITION_INDEPENDENT_CODE target property is set to ON automatically.

So this is good. Sorry for the noise.

ahoarau:
@jmirabel Sorry didnt read it through. Done !

jmirabel (approved):

hungpham2511:
Thanks!

PR #179, opened 2021-08-08
Fixes for windows builds

ahoarau:
Minor PR to make the cpp part compiles on windows64 (msvc).

### DIFF
--- a/cpp/src/toppra/geometric_path/piecewise_poly_path.cpp
+++ b/cpp/src/toppra/geometric_path/piecewise_poly_path.cpp
@@ -153,7 +153,7 @@ size_t PiecewisePolyPath::findSegmentIndex(value_type pos) const {
   }
   auto it = std::upper_bound(m_breakpoints.begin(), m_breakpoints.end(), pos);
   auto idx = std::distance(m_breakpoints.begin(), it)-1;
-  return std::min<size_t>(std::max<size_t>(idx, 0), m_coefficients.size() - 1);
+  return std::min<size_t>(std::max<size_t>(idx, decltype(idx)(0) ), m_coefficients.size() - 1);
 }
 
 void PiecewisePolyPath::checkInputArgs() {
--- a/cpp/src/toppra/geometric_path/piecewise_poly_path.cpp
+++ b/cpp/src/toppra/geometric_path/piecewise_poly_path.cpp
@@ -3,6 +3,7 @@
 #include <toppra/geometric_path/piecewise_poly_path.hpp>
 #include <toppra/toppra.hpp>
 #include <Eigen/Dense>
+#include <array>
 
 #ifdef TOPPRA_OPT_MSGPACK
 #include <msgpack.hpp>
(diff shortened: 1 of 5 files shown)


cpp/src/toppra/geometric_path/piecewise_poly_path.cpp
```diff
@@ -152,7 +153,7 @@ size_t PiecewisePolyPath::findSegmentIndex(value_type pos) const {
   }
   auto it = std::upper_bound(m_breakpoints.begin(), m_breakpoints.end(), pos);
   auto idx = std::distance(m_breakpoints.begin(), it)-1;
-  return std::min(static_cast<size_t>(std::max(idx, long{0})), m_coefficients.size() - 1);
+  return std::min<size_t>(std::max<size_t>(idx, 0), m_coefficients.size() - 1);
```
ahoarau: This one I'm not so sure @hungpham2511

hungpham2511: If the tests are passing I suppose it's fine. @jmirabel would you want to have a look?

jmirabel: Make `idx` a `std::size_t` rather than `auto`. Then, I think on Windows, you have to put `std::min` and `std::max` between parenthesis to avoid min and max being treated as macros (not 100% sure I remember correctly the reason.

Then, this should do the work.
```suggestion
  return (std::min)((std::max)(idx, std::size_t(0)), m_coefficients.size() - 1);
```

ahoarau: The `long{0}` confuses thecompiler if `idx` is `std::size_t`. Im concerned about some potential overflows when dealing with signed /unsigned.

jmirabel: Sorry for coming back to this so late. I did not see your comment.

std::size_t is a `unsigned long`. Overflowing here means that you would have a very large number of coefficients, too large to fit into memory.

The option with no casting is
```suggestion
  return std::min<size_t>(std::max<size_t>(idx, decltype(idx)(0) ), m_coefficients.size() - 1);
```

You can also split the line:
```suggestion
  if (idx <= 0) return 0;
  if (idx >= m_coefficients.size() - 1) return m_coefficients.size() - 1;
  return idx;
```

ahoarau:
I took the liberty of fixing all windows warnings. Tests are running for the cpp part

jmirabel (approved):

jmirabel:
@hungpham2511 CI error looks unrelated to this change.

hungpham2511:
Yep, it's not related. Thank you both for your contribution.

hungpham2511:
@ahoarau could you help to add a line in the [changelog](https://github.com/hungpham2511/toppra/blob/2cc324498237f0185c58f16844bc3fa21f358c50/HISTORY.md)? Then I will happily merge the PR. Thanks again.

PR #185, opened 2021-11-05
Fixed bug in ConstAccel parametrizer where evaluating valid times can sometimes result in an exception being thrown.

### LINKED ISSUE #184: Spurious out of bounds issue in the const accel parametrizer due to floating point imprecision
**Describe the bug**
When the upper bound of ConstAccel::pathInterval_impl() is passed to ConstAccel::eval_impl(), this can sometimes result in a call to the source path's eval() method with an out-of-bounds value. If this path is a PiecewisePolyPath, this results in an exception being thrown. It appears floating-point imprecision is the issue.

**To Reproduce**

1. Run the following snippet
```cpp
toppra::Vectors positions = {
  toppra::Vector::Zero(1),
  toppra::Vector::Zero(1)
};
toppra::Vectors velocities = {
  toppra::Vector::Zero(1),
  toppra::Vector::Zero(1)
};

double PATH_LEN = 25363210.115759313106536865234375;

toppra::Vector times = toppra::Vector::LinSpaced(2, 0, PATH_LEN);
auto std_times = std::vector<double>(times.data(), times.data() + times.size());

auto path = toppra::PiecewisePolyPath::constructHermite(positions, velocities, std_times);
auto ppath = std::make_shared<toppra::PiecewisePolyPath>(path);

toppra::Vector gridpoints = toppra::Vector::LinSpaced(10, 0, PATH_LEN);
toppra::Vector vsquared{10};
vsquared << 0, 0.1, 0.2, 0.3, 0.5, 0.5, 0.3, 0.2, 0.1, 0.0;
auto p = toppra::parametrizer::ConstAccel(ppath, gridpoints, vsquared);

auto bound = p.pathInterval();
auto qs = p.eval_single(bound[1]);
```

2. This results in an exception being thrown with the following error message.
```
Position 2.53632e+07 is outside of range [ 0, 2.53632e+07]
```
...from piecewise_poly_path.cpp:152

**Expected behavior**
The path should not throw an exception when evaluated with times which are within the path's reported interval.

**Version**
Tip of develop - commit: a70f6a217ddf20c42b62471ab52ce16884a320c1

Note: The spline parametrizer doesn't appear to suffer from the same issue.

stevegolton:
Fixes #184

Changes in this PR:
- Fixed bug where the `parametrizer::ConstAccel::eval_single()` can sometimes throw an out-of-bounds error, even when the time passed to it is supposedly within bounds as reported by `parametrizer::ConstAccel::pathInterval()`.

Checklists:

## Description of the PR

Consider the following snippet:
```cpp
auto p = toppra::parametrizer::ConstAccel(path, gridpoints, vsquared);
auto bound = p.pathInterval();
p.eval_single(bound[1]));
```
The call to `eval_single(bound[1])` can sometimes throw an error for certain paths, even though the time `bound[1]` should be a valid evaluation time. This is due to a bug inside the `ConstAccel` parametrizer. This PR fixes this bug.

The `ConstAccel` object contains a pointer to a `GeometricPath` object called `m_path` which is injected in at construction. When `ConstAccel::eval_single()` or `ConstAccel::eval()` is called, the `ConstAccel` object translates the time passed in to a value of `s` which is subsequently used in a call to the underlying `m_path->eval()`.

As long as the time passed into `ConstAccel::eval()` in the first place is within the bounds defined by `ConstAccel::pathInterval()`, the resulting call to `m_path->eval()` should be within the bounds defined by `m_path->pathInterval()`, but due to floating point numerical errors and the way the translation works, the isn't always the case. Sometimes the resulting value of `s` passed to `m_path->eval()` can be outside the bounds defined by `m_path->pathInterval()`, and if `m_path` is of type `PiecewisePolyPath`, a `std::runtime_error` is thrown:

https://github.com/hungpham2511/toppra/blob/a70f6a217ddf20c42b62471ab52ce16884a320c1/cpp/src/toppra/geometric_path/piecewise_poly_path.cpp#L148-L153

In practice, this means it's dangerous to sample arbitrary trajectories using e.g. Eigen's `LinSpaced()` function, as sometimes this will fail when the underlying `m_path` object throws an error, especially with times which are very close to the bounds.
```cpp
auto bound = p.pathInterval();
toppra::Vector times = toppra::Vector::LinSpaced(N, bound[0], bound[1]);
p.eval(times); // <-- can sometimes fail
```

Instead, the caller must do something nasty like this:
```cpp
auto bound = p.pathInterval();
toppra::Vector times = toppra::Vector::LinSpaced(N, bound[0], bound[1]);
const double EPS = 1e-5;
times[0] += EPS;
times[N-1] -= EPS;
p.eval(times); // <-- will hopefully never fail..??!
```

Note: We assume the following assertions hold true, which appears to be the case for the `LinSpaced()` function.
```cpp
toppra::Vector times = toppra::Vector::LinSpaced(N, A, B);
assert(times[0] == A);
assert(times[N-1] == B);
```

The most important thing this PR adds is a clamp on any value before it is passed to the underlying `m_path->eval()` in order to move it within the bounds defined by `m_path->pathInterval()`. However, this means most values passed to `ConstAccel::eval()` which are outside the bounds defined by `ConstAccel::pathInterval()` get silently clamped, which is not friendly for debugging. Thus, the second thing this PR adds is a specific check that all times passed to `ConstAccel::eval()` are within the range `ConstAccel::pathInterval()` otherwise a `std::runtime_error` is thrown, which is exactly what the `PeicewisePolyPath` class does.

This PR does not address any issue which might arise from the caller doing something like the following, either inadvertently or not.
```cpp
auto p = toppra::parametrizer::ConstAccel(path, gridpoints, vsquared);
auto bound = p.pathInterval();
p.eval_single(bound[1] + SOME_TINY_VALUE));
```

This is because the caller can check for this case themselves, whereas in the bug discussed above it is not possible to do so from the outside. Also, this is an obvious violation of the `ConstAccel` object's interface.

### DIFF
--- a/cpp/src/toppra/parametrizer/const_accel.cpp
+++ b/cpp/src/toppra/parametrizer/const_accel.cpp
@@ -61,6 +61,13 @@ Vectors ConstAccel::eval_impl(const Vector& times, int order) const {
   Vector ss, vs, us;
   TOPPRA_LOG_DEBUG("eval_impl. order=" << order);
   bool ret = evalParams(times, ss, vs, us);
+  auto path_interval = m_path->pathInterval();
+  // Clamp all elements of ss to within the bounds of the path
+  for (std::size_t i = 0; i < ss.size(); i++) {
+    if (ss[i] > path_interval[1]) {
+      ss[i] = path_interval[1];
+    }
+  }
   assert(ret);
   switch (order) {
     case 0:
--- a/cpp/src/toppra/parametrizer/const_accel.cpp
+++ b/cpp/src/toppra/parametrizer/const_accel.cpp
@@ -61,13 +61,25 @@ Vectors ConstAccel::eval_impl(const Vector& times, int order) const {
   Vector ss, vs, us;
   TOPPRA_LOG_DEBUG("eval_impl. order=" << order);
   bool ret = evalParams(times, ss, vs, us);
-  auto path_interval = m_path->pathInterval();
-  // Clamp all elements of ss to within the bounds of the path
-  for (std::size_t i = 0; i < ss.size(); i++) {
-    if (ss[i] > path_interval[1]) {
-      ss[i] = path_interval[1];
+
+  auto interval = pathInterval_impl();
+  for (std::size_t i = 0; i < times.size(); ++i)
+  {
+    if ((times[i] < interval[0]) || (times[i] > interval[1]))
+    {
+      std::ostringstream oss;
+      oss << "Time " << times[i] << " is outside of range [ " << interval[0]
+        << ", " << interval[1] << ']';
+      throw std::runtime_error(oss.str());
     }
   }
+
+  // Even if the times are within the bounds of the parametrizer, numerical
+  // errors can still push our values outside the bounds of the path, so
+  // we clamp values of ss to the path interval
+  auto path_interval = m_path->pathInterval();
+  ss = ss.cwiseMax(path_interval[0]).cwiseMin(path_interval[1]);
+
   assert(ret);
   switch (order) {
     case 0:
--- a/cpp/tests/test_parametrizer.cpp
+++ b/cpp/tests/test_parametrizer.cpp
@@ -100,3 +100,30 @@ TEST_F(ParametrizeConstAccel, Correctness) {
     }
   }
 }
+
+TEST(ParametrizeConstAccelNoFixture, BoundsViolation) {
+    toppra::Vectors positions = {
+      toppra::Vector::Zero(1),
+      toppra::Vector::Zero(1)
+    };
+    toppra::Vectors velocities = {
+      toppra::Vector::Zero(1),
+      toppra::Vector::Zero(1)
+    };
+
+    double PATH_LEN = 25363210.115759313106536865234375;
+
+    toppra::Vector times = toppra::Vector::LinSpaced(2, 0, PATH_LEN);
+    auto std_times = std::vector<double>(times.data(), times.data() + times.size());
+
+    auto path = toppra::PiecewisePolyPath::constructHermite(positions, velocities, std_times);
+    auto ppath = std::make_shared<toppra::PiecewisePolyPath>(path);
+
+    toppra::Vector gridpoints = toppra::Vector::LinSpaced(10, 0, PATH_LEN);
+    toppra::Vector vsquared{10};
+    vsquared << 0, 0.1, 0.2, 0.3, 0.5, 0.5, 0.3, 0.2, 0.1, 0.0;
+    auto p = toppra::parametrizer::ConstAccel(ppath, gridpoints, vsquared);
+
+    auto bound = p.pathInterval();
+    auto qs = p.eval_single(bound[1]);
+}
--- a/cpp/tests/test_parametrizer.cpp
+++ b/cpp/tests/test_parametrizer.cpp
@@ -125,5 +125,10 @@ TEST(ParametrizeConstAccelNoFixture, BoundsViolation) {
     auto p = toppra::parametrizer::ConstAccel(ppath, gridpoints, vsquared);
 
     auto bound = p.pathInterval();
-    auto qs = p.eval_single(bound[1]);
+
+    const double EPS = 1e-8;
+    EXPECT_NO_THROW(p.eval_single(bound[0]));
+    EXPECT_THROW(p.eval_single(bound[0] - EPS), std::runtime_error);
+    EXPECT_NO_THROW(p.eval_single(bound[1]));
+    EXPECT_THROW(p.eval_single(bound[1] + EPS), std::runtime_error);
 }
--- a/cpp/tests/test_parametrizer.cpp
+++ b/cpp/tests/test_parametrizer.cpp
@@ -111,7 +111,7 @@ TEST(ParametrizeConstAccelNoFixture, BoundsViolation) {
       toppra::Vector::Zero(1)
     };
 
-    double PATH_LEN = 25363210.115759313106536865234375;
+    double PATH_LEN = 3.51363644474459757560680;
 
     toppra::Vector times = toppra::Vector::LinSpaced(2, 0, PATH_LEN);
     auto std_times = std::vector<double>(times.data(), times.data() + times.size());
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -3,6 +3,7 @@
 ## Unreleased
 
 - [cpp] Minor PR to make the cpp part compiles on windows64 (msvc). Thanks @ahoarau.
+- [cpp] Fixed out of bounds issue in parametrizer::ConstAccel.
 - [python] Fix linting error that causes CI failure
 - [ci] Do integration testing in gh action instead of circle CI
 


hungpham2511:
Thank you for the pull request!

I will try to review and merge as soon as I can. Thanks again for the contribution.

cpp/tests/test_parametrizer.cpp
```diff
@@ -100,3 +100,30 @@ TEST_F(ParametrizeConstAccel, Correctness) {
...
+    auto path = toppra::PiecewisePolyPath::constructHermite(positions, velocities, std_times);
+    auto ppath = std::make_shared<toppra::PiecewisePolyPath>(path);
+
+    toppra::Vector gridpoints = toppra::Vector::LinSpaced(10, 0, PATH_LEN);
+    toppra::Vector vsquared{10};
+    vsquared << 0, 0.1, 0.2, 0.3, 0.5, 0.5, 0.3, 0.2, 0.1, 0.0;
+    auto p = toppra::parametrizer::ConstAccel(ppath, gridpoints, vsquared);
+
+    auto bound = p.pathInterval();
+    auto qs = p.eval_single(bound[1]);
```
jmirabel: To make the test clearer:
```suggestion
    EXPECT_NO_THROW(p.eval_single(bound[1]));
```

cpp/src/toppra/parametrizer/const_accel.cpp
```diff
@@ -61,6 +61,13 @@ Vectors ConstAccel::eval_impl(const Vector& times, int order) const {
   Vector ss, vs, us;
   TOPPRA_LOG_DEBUG("eval_impl. order=" << order);
   bool ret = evalParams(times, ss, vs, us);
+  auto path_interval = m_path->pathInterval();
+  // Clamp all elements of ss to within the bounds of the path
+  for (std::size_t i = 0; i < ss.size(); i++) {
+    if (ss[i] > path_interval[1]) {
+      ss[i] = path_interval[1];
+    }
+  }
```
jmirabel, line 75: I am fine with clamping numerical errors. The way it is done here is non-friendly for debugging because it doesn't fail or even warn when the elements are far above the limit.

stevegolton: Do you think clamping within some epsilon would work here? I'm not at a PC right now but I seem to remember there exists a TOPPRA_EPS which could be used.

jmirabel: You can try with this epsilon. It may not be sufficient in your case since the path length is very big.

jmirabel:
May I know what you try to do ?

If 25363210 is in second, it is equal to 0.8 years, which is non-sense for a robot trajectory.

stevegolton:
> May I know what you try to do ?
> 
> If 25363210 is in second, it is equal to 0.8 years, which is non-sense for a robot trajectory.

I've encountered this bug on sensical trajectories - e.g. trajectories of a few seconds in duration - but I didn't manage to capture the conditions. The test here is just a canned example which happened to reproduce the issue.

Edit:

I just updated the test to use a more realistic path length:
```
double PATH_LEN = 3.51363644474459757560680;
```

cpp/src/toppra/parametrizer/const_accel.cpp
stevegolton: The large path is just a canned example (see comment below).

Another option might be to throw early if any element in times is outside the constaccel bounds, similar to the check in PiecewisePoly. Then the clamping here would only be done if a numerical error has moved a seemingly valid time outside of the bounds of the source path.

Yet another option would be to require a source path that supports extrapolation, but this seems like it would be a surprising requirement.

jmirabel: I am fine with clamping. You can do something like:
```
if ((ss.array() < path_interval[0] * (1 - EPS)).any())
  throw std::runtime_error("a meaningful message");
if ((ss.array() > path_interval[1] * (1 + EPS)).any())
  throw std::runtime_error("a meaningful message");
ss = ss.cwiseMax(path_interval[0]).cwiseMin(path_interval|1]);
```
Usually, the best is to mix relative and absolute test. The above test may not work find when 0 (or close to 0) is one bound of the path interval. If the path interval bound is inferior to EPS, then you may keep EPS.

stevegolton: I still feel like that leaves edge cases like the small path time you just mentioned.

The way I see it, the requirements/test should be something like this:
```cpp
const double EPS = [something really small];
auto bound = p.pathInterval();

// These should never fail for any input path
EXPECT_NO_THROW(p.eval_single(bound[0]));
EXPECT_NO_THROW(p.eval_single(bound[1]));

// These should always fail
EXPECT_THROW(p.eval_single(bound[0] - EPS), std::runtime_error);
EXPECT_THROW(p.eval_single(bound[1] + EPS), std::runtime_error);
```

Thus, if you're not adverse to the extra check then I suggest something like this should be suitable?
```cpp
auto interval = pathInterval_impl();
if ((times.array() < interval[0]).any())
  throw std::runtime_error("a meaningful message");
if ((times.array() > interval[1]).any())
  throw std::runtime_error("a meaningful message");

auto path_interval = m_path->pathInterval();
ss = ss.cwiseMax(path_interval[0]).cwiseMin(path_interval[1]);
```

What do you think?
Also, thanks for the tip about Eigen reduction operators.

hungpham2511:
> > May I know what you try to do ?
> > If 25363210 is in second, it is equal to 0.8 years, which is non-sense for a robot trajectory.
> 
> I've encountered this bug on sensical trajectories - e.g. trajectories of a few seconds in duration - but I didn't manage to capture the conditions. The test here is just a canned example which happened to reproduce the issue.

Sounds interesting. Please let us know if you spot another instance of this error again. This can happen if you have a near-zero path velocity, which again can happen if you i) use the basic interpolation scheme and ii) have a tricky path.

cpp/src/toppra/parametrizer/const_accel.cpp
hungpham2511: I think what @jmirabel means is to use `EPS_REL` and `EPS_ABS` in the check, i.e., 
```
if ((ss.array() < (path_interval[0] * (1 - EPS_REL) - EPS_ABS)).any())
  throw std::runtime_error("a meaningful message");
```

`EPS_ABS` can be given some meaningful values such as 0.1ms (1-e4).

stevegolton:
> > > May I know what you try to do ?
> > > If 25363210 is in second, it is equal to 0.8 years, which is non-sense for a robot trajectory.
> > 
> > 
> > I've encountered this bug on sensical trajectories - e.g. trajectories of a few seconds in duration - but I didn't manage to capture the conditions. The test here is just a canned example which happened to reproduce the issue.
> 
> Sounds interesting. Please let us know if you spot another instance of this error again. This can happen if you have a near-zero path velocity, which again can happen if you i) use the basic interpolation scheme and ii) have a tricky path.

It's quite easy to reproduce by trying random path lengths. Consider the following test, which tries up to 100 random path lengths on the ConstAccel parametrizer. On my machine, I usually end up with this bug cropping up within a handful of attempts. Note: I didn't include this test in the pull request, as randomness doesn't make for a good unit test.

The bug appears to be an artifact of trying to arrive at the same number via different methods which ends up with slightly different values due to floating point numerical errors. This is fine in the middle of the path but breaks down at the two ends of the path.

```cpp
TEST(ParametrizeConstAccelNoFixture, BoundsViolationRandom) {
  std::random_device rd;
  std::mt19937 e2(rd());
  std::uniform_real_distribution<> dist(0, 10.0);

  for (int i = 0; i < 1e3; ++i)
  {
    double PATH_LEN = dist(e2);
    std::cout << std::setprecision(32) << PATH_LEN << '\n';

    toppra::Vectors positions = {
      toppra::Vector::Zero(1),
      toppra::Vector::Zero(1)
    };
    toppra::Vectors velocities = {
      toppra::Vector::Zero(1),
      toppra::Vector::Zero(1)
    };

    toppra::Vector times = toppra::Vector::LinSpaced(2, 0, PATH_LEN);
    auto std_times = std::vector<double>(times.data(), times.data() + times.size());

    auto path = toppra::PiecewisePolyPath::constructHermite(positions, velocities, std_times);
    auto ppath = std::make_shared<toppra::PiecewisePolyPath>(path);

    toppra::Vector gridpoints = toppra::Vector::LinSpaced(10, 0, PATH_LEN);
    toppra::Vector vsquared{10};
    vsquared << 0, 0.1, 0.2, 0.3, 0.5, 0.5, 0.3, 0.2, 0.1, 0.0;
    auto p = toppra::parametrizer::ConstAccel(ppath, gridpoints, vsquared);

    auto bound = p.pathInterval();
    auto qs = p.eval_single(bound[1]);
  }
}
```

cpp/src/toppra/parametrizer/const_accel.cpp
stevegolton: I don't quite understand the advantage of doing it this way. Surely, if a user asks for a number inside `ConstAccel::pathInterval()`, this should always work, and outside should always fail (with a meaningful message). It seems that EPS doesn't come into it at all.

hungpham2511: The sort of scenarios that this code can help catch is numerical errors. Say for example, suppose you generate several time instances that should span [0, 3.1423231], I would expect that the last time instance can be slight _off_.

It's therefore reasonable to clamp if the numerical checks with eps pass. Otherwise, the code should throw exceptions on widely off values (-1 in a [0, 1.0] path).

stevegolton: I feel like it should be up to the caller to make sure their values are exactly within the reported bounds. At least this is the approach taken by the `PiecewisePolyPath` object:

piecewise_poly_path.cpp:148
```cpp
if (pos > m_breakpoints[m_breakpoints.size()-1] || pos < m_breakpoints[0]) {
  std::ostringstream oss;
  oss << "Position " << pos << " is outside of range [ " << m_breakpoints[0]
    << ", " << m_breakpoints[m_breakpoints.size()-1] << ']';
  throw std::runtime_error(oss.str());
}
```
I'm not adverse to the EPS approach, I just don't think it's necessary. Your call though.

jmirabel: > I feel like it should be up to the caller to make sure their values are exactly within the reported bounds.

Isn't it what is implemented in the current HEAD. You get an exception if you are not exactly within bounds. And this is what your PR tries to address...

relative + absolute check make sense.

stevegolton: That's right, the current HEAD fixes my issue and the test I have written asserts this. I evidently haven't done a good job at explaining what this PR is trying to address, so I'll rename my PR for clarity. I'm not trying to protect the caller from themselves, rather I'm fixing a bug inside the `ConstAccel` parametrizer.

In a nutshell, I hope you agree that this snippet of code should never fail, assuming `p` is a valid `ConstAccel` parametrizer object.
```cpp
auto bound = p.pathInterval();
auto qs = p.eval_single(bound[1]); // <-- this should never throw an error, but it sometimes does on the current tip of develop
```

stevegolton: I've added a comprehensive description of the bug above, and my reasoning for the content of this PR. I hope this helps clear up any confusion :D

Let me know if you have any questions.

hungpham2511:
> > > > May I know what you try to do ?
> > > > If 25363210 is in second, it is equal to 0.8 years, which is non-sense for a robot trajectory.
> > > 
> > > 
> > > I've encountered this bug on sensical trajectories - e.g. trajectories of a few seconds in duration - but I didn't manage to capture the conditions. The test here is just a canned example which happened to reproduce the issue.
> > 
> > 
> > Sounds interesting. Please let us know if you spot another instance of this error again. This can happen if you have a near-zero path velocity, which again can happen if you i) use the basic interpolation scheme and ii) have a tricky path.
> 
> It's quite easy to reproduce by trying random path lengths. Consider the following test, which tries up to 100 random path lengths on the ConstAccel parametrizer. On my machine, I usually end up with this bug cropping up within a handful of attempts. Note: I didn't include this test in the pull request, as randomness doesn't make for a good unit test.
> 
> The bug appears to be an artifact of trying to arrive at the same number via different methods which ends up with slightly different values due to floating point numerical errors. This is fine in the middle of the path but breaks down at the two ends of the path.
> 
> ```c++
> TEST(ParametrizeConstAccelNoFixture, BoundsViolationRandom) {
>   std::random_device rd;
>   std::mt19937 e2(rd());
>   std::uniform_real_distribution<> dist(0, 10.0);
> 
>   for (int i = 0; i < 1e3; ++i)
>   {
>     double PATH_LEN = dist(e2);
>     std::cout << std::setprecision(32) << PATH_LEN << '\n';
> 
>     toppra::Vectors positions = {
>       toppra::Vector::Zero(1),
>       toppra::Vector::Zero(1)
>     };
>     toppra::Vectors velocities = {
>       toppra::Vector::Zero(1),
>       toppra::Vector::Zero(1)
>     };
> 
>     toppra::Vector times = toppra::Vector::LinSpaced(2, 0, PATH_LEN);
>     auto std_times = std::vector<double>(times.data(), times.data() + times.size());
> 
>     auto path = toppra::PiecewisePolyPath::constructHermite(positions, velocities, std_times);
>     auto ppath = std::make_shared<toppra::PiecewisePolyPath>(path);
> 
>     toppra::Vector gridpoints = toppra::Vector::LinSpaced(10, 0, PATH_LEN);
>     toppra::Vector vsquared{10};
>     vsquared << 0, 0.1, 0.2, 0.3, 0.5, 0.5, 0.3, 0.2, 0.1, 0.0;
>     auto p = toppra::parametrizer::ConstAccel(ppath, gridpoints, vsquared);
> 
>     auto bound = p.pathInterval();
>     auto qs = p.eval_single(bound[1]);
>   }
> }
> ```

This certainly sounds like a bug :sweat_smile:

cpp/src/toppra/parametrizer/const_accel.cpp
hungpham2511: Thanks for the edits, it's clear for me now. 

However I think we can allow the user to provide slightly out-of-bound `times`, so that small numerical errors are acceptable (1e-10 level). In particular, 
```cpp
auto interval = pathInterval_impl();
if ((times.array() < (interval[0] * (1 - EPS_REL) - EPS_ABS)).any())
  throw std::runtime_error("a meaningful message");
// similar to the upper bound
```

The clamping can be kept as is, as we know "mathematically" that `ss` is now valid if no exceptions were raised.

hungpham2511: Hi @stevegolton, gentle ping. Do you still need this PR to be merged?

stevegolton: Hey @hungpham2511, apologies, I got a little stuck on the updates you suggested above.

I'm happy with the PR as-is because it fixes the bug. I understand what EPS_REL and EPS_ABS is for, but I don't understand what problem it is trying to solve, so I don't know how to effectively test it or what values to choose.

The only example I can think of, a naive usage of `ConstAccel` might be something like below. Is this what you want to protect against?

```cpp
int SEGCOUNT = 10;
auto pathlen = p.pathInterval()[1] - p.pathInterval()[0];
auto seglen = pathlen / SEGCOUNT;
for (int i = 0; i <= SEGCOUNT; i++) {
  auto t = p.pathInterval()[0] + i * seglen; // <-- t might be > p.pathInterval()[1] when i == SEGCOUNT
  p.eval_single(t);
}
```

Please do what you think best with this PR, feel free to close it or add the REL/ABS changes yourself if you think it's worth keeping the bugfix. Might I suggest you merge this PR, and we address the REL/ABS issue in a separate PR?

hungpham2511: Hi @stevegolton, thanks for the work anyway. I think I understand your blocking point so I will work on this in the next few days to get it to merge to develop. Thanks again for the work.

stevegolton: Thanks @hungpham2511 for all your help. I'm happy with how this turned out.

PR #190, opened 2021-11-12
Fixes #189. Fixed incorrect variable bug in seidel.cpp which was causing memory issues.

### LINKED ISSUE #189: Incorrect variable used in seidel.cpp, causing memory errors
**Describe the bug**
It appears a loop counter variable `i` shadows another loop counter variable with the same name, used to iterate an outer loop. This would be fine, but the outer loop variable is used in the scope of the inner loop as well, so the wrong variable is being used erroneously. This most likely results in the wrong behaviour, but can also occasionally result in reading off the end of the vector and unallocated memory being accessed, which is reported by valgrind as an invalid memory access. This usually has no immediate effect but can cause crashes further down the line, especially when memory is being freed.
https://github.com/hungpham2511/toppra/blob/a70f6a217ddf20c42b62471ab52ce16884a320c1/cpp/src/toppra/solver/seidel.cpp#L326-L327

I'm not 100% sure of the solution, but I think it should probably look something like this instead:
```
for (int k = 0; k < m_A[i].rows(); ++k)
      m_index_map[k] = k;
```

**Version**
Tip of develop.

stevegolton:
Fixes #189.

Changes in this PRs:
- Fixed incorrect variable usage in the seidel solver (see #189).

Checklists:

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -3,6 +3,7 @@
 ## Unreleased
 
 - [cpp] Minor PR to make the cpp part compiles on windows64 (msvc). Thanks @ahoarau.
+- [cpp] Fixed incorrect variable bug in seidel solver.
 - [python] Fix linting error that causes CI failure
 - [ci] Do integration testing in gh action instead of circle CI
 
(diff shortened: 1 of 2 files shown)


jmirabel (approved):

hungpham2511:
Thanks.

PR #192, opened 2021-11-19
Google Test: master -> main

compscidr:
https://github.com/google/googletest - doesn't seem to have a master branch but a main branch now

There isn't an active issue, but noticed one of our CI jobs failing this morning and I believe this is the cause.

PS: you may want to cherry-pick back into previous tags because I believe they may also fail to build for the same reason.


hungpham2511:
Thanks for the PR. Indeed develop branch is failing.

PR #208, opened 2022-08-19
fix the f-string bug when hot_qpoases solver failed

jsbyysheng:
This pr will fix the f-string bug for `TypeError: not all arguments converted during string formatting` and deprecation warning for `The 'warn' method is deprecated, use 'warning' instead`. I will update HISTORY.md if you agree with this pr.

### DIFF
--- a/toppra/solverwrapper/hot_qpoases_solverwrapper.py
+++ b/toppra/solverwrapper/hot_qpoases_solverwrapper.py
@@ -269,7 +269,8 @@ class hotqpOASESSolverWrapper(SolverWrapper):
             ):
                 logger.fatal(
                     f"(0, 0) satisfies all constraints => error due to numerical errors, \
-                        {self._A}, {self._lA}, {self._hA}, {self._l}, {self._h}"
+                        %s, %s, %s, %s, %s",  self._A, self._lA, self._hA, self._l, self._h
+
                 )
             else:
                 logger.debug("(0, 0) does not satisfy all constraints.")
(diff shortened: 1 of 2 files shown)


toppra/solverwrapper/hot_qpoases_solverwrapper.py
```diff
@@ -268,12 +268,8 @@ def solve_stagewise_optim(self, i, H, g, x_min, x_max, x_next_min, x_next_max):
...
             ):
                 logger.fatal(
-                    "(0, 0) satisfies all constraints => error due to numerical errors.",
-                    self._A,
-                    self._lA,
-                    self._hA,
-                    self._l,
-                    self._h,
+                    f"(0, 0) satisfies all constraints => error due to numerical errors, \
+                        {self._A}, {self._lA}, {self._hA}, {self._l}, {self._h}"
```
hungpham2511:
```suggestion
                    f"(0, 0) satisfies all constraints => error due to numerical errors, \
                        %s, %s, %s, %s, %s",  self._A, self._lA, self._hA, self._l, self._h

```

This would be slightly more efficient as it allows Python to do the string interpolation later.

hungpham2511 (commented):
Thanks for your contribution!

Both LGTMs, can you modify the logging statement as suggested?

jsbyysheng:
Glad to hear from you. I have accepted your suggestion.

hungpham2511:
> Glad to hear from you. I have accepted your suggestion.

Thanks and sorry for the delay.

PR #215, opened 2022-11-18
chore: make C++ always define all installed symbols.

jmirabel:
Avoid having undefined symbols when the library is not compiled with GLPK or qpOASES support.

Checklists:

After this is merge, @hungpham2511 could you make a patch release pls ?


hungpham2511:
@jmirabel 0.5.2 has been released with a new tag.

PR #223, opened 2023-02-12
[Feature] seidel solver(cython) solves 1D LP when possible.

ompugao:
Fixes:
https://github.com/hungpham2511/toppra/blob/a35f2de47087d962224ba9ffc73501c7b22ca238/toppra/solverwrapper/cy_seidel_solverwrapper.pyx#L549

Changes in this PRs:
- if solve_lp1d flag is set to > 0 and x_min == x_max, seidel solver (cython version) solves an 1D LP instead of a 2D LP.

Checklists:

### DIFF
--- a/toppra/solverwrapper/cy_seidel_solverwrapper.pyx
+++ b/toppra/solverwrapper/cy_seidel_solverwrapper.pyx
@@ -416,6 +416,9 @@ cdef class seidelWrapper:
         double [:, ::1] a_arr, b_arr, c_arr  # mmviews of coefficients of the 2D Lp
         double [:, :] low_arr, high_arr    # mmviews of coefficients of the 2D Lp
         double scaling # path scaling
+
+        double [:] bx_c  # mmviews of coefficients of the 1D Lp
+
         
     # @cython.profile(True)
     def __init__(self, list constraint_list, path, path_discretization):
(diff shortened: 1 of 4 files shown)


hungpham2511 (commented):
Do you have any result on the performance gain from this optimization? 

I think it would be rather minimal but I might be wrong.

ompugao:
I got roughly 1.2~1.3x better performance for forward_step.
This would be negligible for usual cases such as topp since other parts are the bottleneck,
But I call self._forward_step many times (~400k) in my project and this change leads to a better performance.

toppra/solverwrapper/cy_seidel_solverwrapper.pyx
```diff
@@ -432,13 +436,17 @@ cdef class seidelWrapper:
             The geometric path.
         path_discretization: array
             The discretized path positions.
+        solve_lp1d: int
+            if solve_lp1d > 0, Solve 1D LP instead of 2D LP when possible (x_min == x_max)
```
hungpham2511:
```suggestion
            if solve_lp1d > 0, solve 1D LP instead of the full 2D LP when possible: when x_min == x_max, solve for u as the variable.
```

hungpham2511 (approved):
The code LGTM, I will merge once CI is happy.

hungpham2511:
Hi, look like CI is having some issue on develop so the failure is not due to your PR. I will try to find some time to fix it next week.

hungpham2511:
Thanks.

PR #224, opened 2023-02-16
Fix planning_utils.py: toppra.compute_trajectory() does not return aux data.

ompugao:
Fixes:
planning_utils.py does not follow the latest api of toppra.algorithm.
TOPPRA instance does not return aux data.

Changes in this PRs:
- not to receive a aux value in planning_utils.py

Checklists:

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -2,6 +2,7 @@
 
 ### Added
 - [python] seidelWrapper solves 1d LP instead of 2d LP when x_min == x_max and solve_1d > 0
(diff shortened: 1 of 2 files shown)


ompugao:
should I add this to history.md? this script will not be under maintainance and i am not sure if you want to keep a record.

hungpham2511:
> should I add this to history.md? this script will not be under maintainance and i am not sure if you want to keep a record.

Sure, please add to changelog. And thanks for the contribution.

PR #232, opened 2023-04-21
fix: fix numerical issue in Seidel 1D solver.

jmirabel:
Rework handling of numerical issues to a more theoretically sound method.

Checklists:

### DIFF
--- a/cpp/src/toppra/solver/seidel-internal.hpp
+++ b/cpp/src/toppra/solver/seidel-internal.hpp
@@ -60,6 +60,8 @@ constexpr value_type infinity = std::numeric_limits<value_type>::infinity();
 
 #define TOPPRA_SEIDEL_LP1D(w,X)                         \
   TOPPRA_LOG_##w("Seidel LP 1D:\n"                      \
+      << "max  v * [ x, 1 ]\n"                          \
+      << "s.t. A * [ x, 1 ] <= 0\n"                     \
       << "v: " << v << '\n'                             \
       << "A:\n" << A << '\n'                            \
       << X);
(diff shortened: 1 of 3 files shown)


cpp/src/toppra/solver/seidel-internal.hpp
```diff
@@ -84,19 +86,24 @@ LpSol1d solve_lp1d(const RowVector2& v, const Eigen::MatrixBase<Derived>& A)
...
+    // a[i] -> a and b[i] -> b in this comment
+    //
+    // case a == 0.0:
+    //   feasible iif b < ABS_TOLERANCE (tolerate a small violation)
+    // otherwise
+    //   if a>0 b<0 then x <= +|b/a| so handled as a normal constraint
+    //   if a<0 b<0 then x >= -|b/a| so handled as a normal constraint
+    //
+    //   if a>0 b>0 then x <= -|b/a| so infeasible if |b/a| > 1/REL_TOLERANCE
+    //   if a<0 b>0 then x >= +|b/a| so infeasible if |b/a| > 1/REL_TOLERANCE
```
jmirabel, line 99: See this comment in the code to know how numerical issues are now handled.

hungpham2511: It's implicit (to me at least) that x is not allowed to be outside of - 1 / REL and 1 / REL range, yes?

If so please document it.

hungpham2511: Otherwise the logic looks fine to me.

hungpham2511 (approved):
LGTM

cpp/src/toppra/solver/seidel-internal.hpp
jmirabel: The fact that it is not allowed is a consequence of the logic but we can indeed state it.

jmirabel: Actually, no, `x` can be outside the range you say.
1. When `b<0`, we do not change the constraint.
2. When `b>0`, the problem is said infeasible if `|b/a|` is too large. But if it is not, the constraint is only on one side of `x` and the optimum may actually be on the other direction, which may be bound beyond +/- 1/REL.

PR #265, opened 2025-10-03
Use ASSERT_NEAR for 0 comparison in test_parametrizer.cpp

### LINKED ISSUE #264: test_parametrizer.cpp fails due to numerical precision error
**Describe the bug**
test_parametrizer.cpp fails due to numerical precision errors.

**To Reproduce**
This is intermittent on some machines, but running the tests on my dev machine produces a failure [here](https://github.com/hungpham2511/toppra/blob/develop/cpp/tests/test_parametrizer.cpp#L51):

```Failure
    Expected equality of these values:
      qds[9][0]
        Which is: 7.019294709056063e-17
      0
```

**Expected behavior**
Test should pass.

**Version**
`develop`

eholum:
Fixes #264 

I'm not sure if this is paving over a more fundamental issue in interpolation or not, but very, very small rounding errors in fp arithmetic seem to be making the final time point in the basic test non-zero in some cases. I _think_ that the minute values are still ok, so we could use FP comparisons instead of exact comparisons in this test. 

Changes in this PRs:
- Use `ASSERT_NEAR` for 0 comparisons

Checklists:
- [ ] Update HISTORY.md with a single line describing this PR

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -1,5 +1,6 @@
 # Changelog
 
+- [cpp]: fix: Update parametrizer test comparisons (#264)
 - [cpp]: fix: Seidel LP 1D: incoherent bounds (#244)
 - [cpp]: fix: Cannot convert from 'initializer list' to 'toppra::BoundaryCond' (#245)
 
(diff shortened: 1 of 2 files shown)


hungpham2511 (approved):

hungpham2511:
Thanks @eholum

PR #266, opened 2025-11-06
build: Move Python library to its own subfolder to support building as ROS 2 package

sea-bass:
Fixes #

Changes in this PRs:
- Moves the Python library to its own subfolder and adds a `COLCON_IGNORE` to it, to allow this package to build in ROS as is.
- Other small updates as I tried things out on my Ubuntu 24.04 / Python 3.12 / ROS 2 Kilted setup
- Deprecated stuff for inactive Python versions (2 and 3.5)

Checklists:


sea-bass:
@hungpham2511 would you be open to a change like this? It probably doesn't fully work right away but I would need your approval to run the CI jobs and get feedback.

Anyways, it makes the build fine even though the Python bindings seem to segfault when I run the tests, regardless of this fix.

hungpham2511:
Thanks for the suggestion. I appreciate the change but I am not using colcon myself so I won't be able to review the change, meanwhile the python path change might break existing code.

Will try to have a closer look this weekend to see if we can do anything about this.

sea-bass:
Thanks! The only tricky thing w ROS is that colcon automatically picks up `setup.py` files in its build unless you slap a `COLCON_IGNORE` file on it.

But with the current folder structure you cannot do this without also ignoring the build of the c++ library.

See https://github.com/colcon/colcon-core/issues/706 for more info

hungpham2511:
Hi @sea-bass, thank you for the PR. I am open to merging it as is as I am planning to drop support for the Python version anyway. The C++ version is likely less numerically stable, but it should be the way to go forward.

Let me know when you are done with the change.

sea-bass:
I am done with it, it's already working for me as is!

(Well, just the c++ library works, the bindings don't seem to be but I am not currently using those)

PR #259, opened 2025-02-26
Updated interface for numpy > 2.0

zkingston:
Fixes an issue with TOPP-RA's incompatibility with more recent NumPy versions.

Changes in this PRs:
- Update Cython interface for NumPy 2.0

Checklists:
- [ ] Update HISTORY.md with a single line describing this PR

PR #268, opened 2026-01-23
Bump version to 0.6.4

traversaro:
Changes in this PRs:
- Bump version to 0.6.4

Checklists:

Hello @hungpham2511, thanks a lot for the work on toppra! For packaging it in conda-forge (or also for other package managers, I guess) it would be convenient to have a tag (similar to https://github.com/hungpham2511/toppra/releases/tag/v0.6.2) with the changes from the last ~2 years. I prepared this PR to bump the version to 0.6.3, do you think it make sense? Thanks a lot in advance!


sea-bass:
@traversaro @hungpham2511 

I just realized that there already is a 0.6.3 on pypi after https://github.com/hungpham2511/toppra/pull/259 -- just that the repo versions and releases were never updated accordingly. So this would have to be 0.6.4 minimally.

Based on some of my latest PRs that move around a bunch of the code and update to newer versions of Python/C++/Linux, maybe we should consider making this one a 0.7.0 instead?

traversaro:
> @traversaro @hungpham2511
> 
> I just realized that there already is a 0.6.3 on pypi after #259 -- just that the repo versions and releases were never updated accordingly. So this would have to be 0.6.4 minimally.
> 
> Based on some of my latest PRs that move around a bunch of the code and update to newer versions of Python/C++/Linux, maybe we should consider making this one a 0.7.0 instead?

Good point! I used 0.6.3 also here as the only changes since https://github.com/hungpham2511/toppra/pull/259 are: 

* https://github.com/hungpham2511/toppra/pull/265
* https://github.com/hungpham2511/toppra/commit/bc2507d91ed66587d5649af32cac8612dc88a0c9
* https://github.com/hungpham2511/toppra/pull/266

and they did not affected the Python library, but if @hungpham2511 prefer 0.6.4 or 0.7.0 that is perfectly fine for me and I can update the PR.

hungpham2511:
@traversaro Thanks for the PR. 0.6.4 sounds good! Since you already have the PR, could you please bump it to 0.6.4?

I don't intend to further any development on the Python library so keeping it at 0.6.x makes more sense to me personally.

traversaro:
> @traversaro Thanks for the PR. 0.6.4 sounds good! Since you already have the PR, could you please bump it to 0.6.4?
> 
> I don't intend to further any development on the Python library so keeping it at 0.6.x makes more sense to me personally.

Done!

sea-bass:
Thanks! Let's land #270 first and then this?

hungpham2511 (approved):
Thanks!

PR #270, opened 2026-01-24
Fix Python and C++ CI, disable ECOS solver in tests

sea-bass:
Changes in this PRs:
- Updates Python setup to work on latest versions
- Changes linting to happen on Python 3.12
- Fixes qpOASES install -- requires https://github.com/hungpham2511/qpOASES/pull/1
- Disables all ECOS solver tests, they don't work and I have no idea how to fix that at the moment
- Updates C++ standard from C++11 to C++14 for new versions
- Removes Ubuntu 20.04 C++ testing due to EOL / adds Ubuntu 24.04 instead
- Drops support for Python 2 in `tox` tasks, but this has been deprecated for a while

Checklists:

### DIFF
--- a/.github/workflows/integrate.yml
+++ b/.github/workflows/integrate.yml
@@ -9,7 +9,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ["3.8", "3.9", "3.10"]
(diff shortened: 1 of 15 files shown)


hungpham2511 (approved):
Thanks!

PR #288, opened 2026-04-20
[cpp] Fix missing cassert includes in headers

sea-bass:
Changes in this PR:
- #280 did not catch them all, as I just learned the hard way trying some conda stuff... 
- The culprit is that some of the `cassert` includes were in the `.cpp` files when in fact they should have been up at the `.hpp` level.

Apologies for the continued tweaks, but it does take some effort to put packages like these through -- in this case -- several build/packaging workflows and discovering surprises along the way. It will stabilize over time!

Checklists:

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of constraint validation and solver checks by ensuring assertion checks are consistently available across headers and implementation files, reducing risk of missed runtime validations.

* **Documentation**
  * Added an entry to the changelog under Upcoming documenting these fixes for transparency.

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -3,7 +3,7 @@
 ## Unreleased
 
 - [cpp] fix: pinocchio v4 compatibility
-- [cpp] Add missing cassert includes
+- [cpp] Fix missing cassert includes in headers
 
 ## 0.6.7 (April 9 2026)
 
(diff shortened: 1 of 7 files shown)


hungpham2511 (approved):
Tks!

PR #279, opened 2026-03-30
[cpp] Fix CMake syntax for Eigen3 version check

sea-bass:
Changes in this PRs:
- conda builds started failing in the new tag because of this. Seems we were using old variable names.
- I also removed some of the long deprecated conditions on the else/endif

See https://github.com/conda-forge/toppra-feedstock/pull/6

Checklists:
- [x] Update HISTORY.md with a single line describing this PR

## Summary by CodeRabbit

* **Chores**
  * Improved build logic for Eigen version detection to ensure correct include/link behavior across releases.
  * Simplified conditional structure in the build configuration for more reliable compilation.

* **Documentation**
  * Added an "Upcoming" entry to the project history noting the C++ build configuration update.

PR #290, opened 2026-05-14
[cpp] Simplify vector typedefs by removing allocators and drop support for Eigen 3.3 and lower

sea-bass:
Fixes #

Changes in this PR:
- I discovered some ABI stability issues with Eigen allocators when upgrading versions from `conda-forge`, and (admittedly due to the help of Claude) landed on simplifying this as in this PR.
- The root cause is the changes in https://github.com/hungpham2511/toppra/pull/279 apparently exposing some issue that was being hidden by using old-style syntax
- So, this also drops support for Eigen <= 3.3

This is the Claude summary:

```
This is PR #279 (Fix CMake syntax for Eigen3 version check):

  -if (${EIGEN3_VERSION_STRING} VERSION_LESS 3.3)
  +if (Eigen3_VERSION VERSION_LESS "3.3")
     target_compile_definitions(toppra PUBLIC ${EIGEN3_DEFINITIONS})
     target_include_directories(toppra PUBLIC ${EIGEN3_INCLUDE_DIRS})
  -else (${EIGEN3_VERSION_STRING} VERSION_LESS 3.3)
  +else()
     target_link_libraries(toppra PUBLIC Eigen3::Eigen)
  -endif (${EIGEN3_VERSION_STRING} VERSION_LESS 3.3)
  +endif()

  With modern Eigen3-Config.cmake (which is what conda-forge ships), EIGEN3_VERSION_STRING is empty and Eigen3_VERSION is
  "3.4.0". CMake compares empty strings as 0, so:

  - 0.6.4 build: "" VERSION_LESS 3.3 → TRUE → took the if branch → toppra got the legacy Eigen path with
  ${EIGEN3_DEFINITIONS} (empty) and ${EIGEN3_INCLUDE_DIRS} (which Config-style find_package does set as a back-compat
  alias). Crucially, Eigen3::Eigen was never added to target_link_libraries(toppra PUBLIC ...), so the exported
  toppraTargets.cmake had no Eigen interface dependency.
  - 0.6.6 build: Eigen3_VERSION VERSION_LESS "3.3" → FALSE → took the else branch → toppra now properly links PUBLIC 
  Eigen3::Eigen. The exported targets file picks this up — and indeed your installed toppraTargets.cmake shows
  INTERFACE_LINK_LIBRARIES "Eigen3::Eigen".

  Why this would cause your segfault

  The pathInterval() segfault is about an offset within PiecewisePolyPath (specifically m_breakpoints at byte offset 0x58).
   That offset is determined by sizeof of the preceding members — the three Matrices = std::vector<Eigen::Matrix, 
  Eigen::aligned_allocator<Eigen::Matrix>>.

  If libtoppra.so was built against one set of Eigen alignment macros (no aligned-allocator state) and your locally-built
  libroboplan_toppra.so is now compiling <toppra/geometric_path/piecewise_poly_path.hpp> against a different set (via the
  new Eigen3::Eigen interface propagation, which can bring in different include order / preprocessor defines than your
  previous setup), the two sides disagree on where m_breakpoints lives in the object. Local code writes the breakpoints to
  one offset; the disassembled libtoppra.so reads from 0x58 — gets garbage / nullptr / a std::vector whose _M_start is
  nonsense — and crashes. This perfectly explains why eval() and computePathParametrization() work (they go through
  m_coefficients, whose offsets the local code and toppra both compute) while pathInterval() doesn't.

  The 0.6.4 CMake bug was hiding this latent ABI mismatch by not exporting the Eigen interface; 0.6.6 fixed the CMake bug,
  which exposed it.
```

Checklists:

## Summary by CodeRabbit

* **Refactor**
  * Simplified public vector container types to use standard containers and made small internal simplifications for consistency.
* **Documentation**
  * Added an “Upcoming” changelog entry describing the typedef simplification.
* **Tests**
  * Updated a test helper to align with the adjusted vector types.
* **Chores**
  * Updated ignore rules and simplified build configuration wiring.

[Review Change Stack](https://app.coderabbit.ai/change-stack/hungpham2511/toppra/pull/290)


sea-bass:
@hungpham2511 Going to include this in the next patch release. Would you be willing to review?

hungpham2511 (approved):
Tks for the patch! @sea-bass

